PR #27, opened 2017-05-06
Allow plugins to expose a different content type (e.g. text/html)

khenderick:
Example for a plugin's index page, returning `text/html`.

```python
@om_expose(content_type='text/html')
def html_index(self):
    self.logger('Building HTML')
    try:
        data = json.dumps(json.loads(self.webinterface.get_modules(self.webinterface.dummy_token)), indent=4)
    except CommunicationTimedOutException as ex:
        data = ex
        self.logger('Error getting modules: CommunicationTimedOutException')
    except Exception as ex:
        data = ex
        self.logger('Error getting modules: {0}'.format(ex))
    html = '<html><head><title>Example</title></head><body><pre>{0}</pre></body></html>'.format(data)
    self.logger('Return HTML')
    return html
```

Also note the requirement to pass `self.webinterface.dummy_token` as token, instead of `None` elsewhere in the plugins.

### DIFF
--- a/src/plugins/base.py
+++ b/src/plugins/base.py
@@ -34,7 +34,7 @@ from datetime import datetime
 import cherrypy
 
 
-def om_expose(method=None, auth=True):
+def om_expose(method=None, auth=True, content_type='application/json'):
     """ Decorator to expose a method of the plugin class through the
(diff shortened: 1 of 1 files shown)


khenderick:
Added for openmotics/gateway-frontend#76

PR #30, opened 2017-08-10
New metrics system

khenderick:
Introducing the new metrics system.
* The Gateway itself pushes a wide set of metrics to the new metrics system
* Plugins can provide metrics to the metrics system as well
* Plugins can subscribe to receive a (sub)set of metrics once they arrive

This will reduce the need for various plugins to make their own (slow) API calls to fetch various information.


khenderick:
Currently a work in progress.

khenderick:
Introduces the requirement/feature of loading egg files for more flexible dependencies. This is now required for the `ws4py` dependency that will have to be shipped/distributed alongside the Gateway package. This is allowed by its BSD-3 license. More info on `ws4py`: Lawouach/WebSocket-for-Python

PR #118, opened 2018-12-28
Fixed bug where subprocess hangs if a ntpdate time update happens

pilaar:


khenderick:
Nope, reporting doesn't seem to be triggered.

PR #132, opened 2019-02-14
Add state/status to events

khenderick:


src/gateway/observer.py
```diff
@@ -198,12 +199,14 @@ def get_outputs(self):
...
 
-    def _output_changed(self, output_id):
+    def _output_changed(self, output_id, status):
         """ Executed by the Output Status tracker when an output changed state """
         self._dbus_service.send_event(DBusEvents.OUTPUT_CHANGE, {'id': output_id})
         for callback in self._event_subscriptions:
             callback({'type': 'OUTPUT_CHANGE',
-                      'data': {'data': {'id': output_id},
+                      'data': {'data': {'id': output_id,
+                                        'status': {'on': status['on'],
```
svanscho: double wrapping of data elements looks a bit messy, don't know the exact app context but can't the parent element be named more meaningfully like e.g. event or payload? Comment goes for all callbacks btw. Also could be nice to have a callback event/data object with helper functions to complete (and standardize) the objects)?

src/master/outputs.py
```diff
@@ -80,8 +80,9 @@ def _update_maybe_report_change(self, output, new_output):
...
                 report = True
         if report:
-            self._report_change(output['id'])
+            self._report_change(output['id'], status=output['status'], dimmer=output['dimmer'])
 
-    def _report_change(self, output_id):
+    def _report_change(self, output_id, status, dimmer):
         if self._on_output_change is not None:
-            self._on_output_change(output_id)
+            self._on_output_change(output_id, {'on': status == 1,
```
svanscho: Is this the most pythonesc way of converting this to a bool? Thinking of something like bool(status)... in case this value is some other integer like e.g. 2? (I'm sure it's not intended to be 2 but :-) )

src/master/shutters.py
```diff
@@ -124,6 +124,6 @@ def get_states(self):
             all_states.append(self._states[i][1])
         return all_states
 
-    def _report_change(self, shutter_id, shutter_data):
+    def _report_change(self, shutter_id, shutter_data, shutter_state):
         if self._on_shutter_change is not None:
-            self._on_shutter_change(shutter_id, shutter_data)
+            self._on_shutter_change(shutter_id, shutter_data, shutter_state[1].upper())
```
svanscho: except IndexError: ? Not sure where you want to catch this - might be handled in upstream code path.,

src/master/thermostats.py
```diff
@@ -80,13 +83,32 @@ def full_update(self, thermostats):
...
         """ Return the list of Outputs. """
         return self._thermostats
 
-    def _report_change(self, thermostat_id):
-        if self._on_thermostat_change is not None:
-            self._on_thermostat_change(thermostat_id)
+    def _report_change(self, thermostat_id, status):
+        if self._on_thermostat_change is not None and status is not None:
+            self._on_thermostat_change(thermostat_id, {'preset': ThermostatStatus._serialize_preset(status['setpoint']),
+                                                       'current_setpoint': status['csetp'],
```
svanscho: Same comment as above, might be handy to wrap more complex data structures in objects.

src/master/thermostats.py
```diff
@@ -80,13 +83,32 @@ def full_update(self, thermostats):
...
+                                                       'output_0': status['output0'],
+                                                       'output_1': status['output1']})
+
+    def _report_group_change(self, thermostats_on, cooling):
+        if self._on_thermostat_group_change is not None:
+            self._on_thermostat_group_change({'state': 'ON' if thermostats_on else 'OFF',
+                                              'mode': 'COOLING' if cooling else 'HEATING'})
+
+    @staticmethod
+    def _serialize_preset(setpoint):
```
svanscho: Was thinking of an enum, but that would require python 3.x - guess it's ok for now

khenderick: You can still uses class constants which is better than just plain strings. But let's keep that for when thermostats are refactored.

src/master/shutters.py
khenderick: It's an internal method using internal fields always containing a two-element list, so high unlikely that there would be an index error at this point.

src/master/thermostats.py
khenderick: Should be handled when thermostats are refactored.

PR #133, opened 2019-02-18
Make sure only physical modules are being bootloaded

khenderick:
A physical module is defined by upper case type letters whilst their virtual equivalents have a lowercase type letter. The current code compares cases insensitive which causes the update to try to update virtual modules (which can't be updated obviously). The most important part of this update fixes this.

It also make sure that all logging is executed similar, so the output is kept in order. Previously stacktraces would end up in the logging on a wrong location, making it quite hard to debug.

### DIFF
--- a/src/modules_bootloader.py
+++ b/src/modules_bootloader.py
@@ -14,11 +14,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 Tool to bootload the slave modules (output, dimmer, input and temperature).
-
-@author: fryckbos
 """
 
 import argparse
+import sys
+import time
+import traceback
 from ConfigParser import ConfigParser
 from serial import Serial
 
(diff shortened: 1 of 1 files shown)


src/modules_bootloader.py
```diff
@@ -279,37 +305,34 @@ def bootload_modules(type, filename, verbose, logger):
...
-                        required=True, help='the type of module to bootload (choices: o, r, d, i, t, c)')
+    parser.add_argument('-t', '--type', dest='type', choices=['o', 'r', 'd', 'i', 't', 'c', 'O', 'R', 'D', 'I', 'T', 'C'], required=True,
+                        help='the type of module to bootload (choices: O, R, D, I, T, C)')
     parser.add_argument('-f', '--file', dest='file', required=True,
                         help='the filename of the hex file to bootload')
-
     parser.add_argument('-l', '--log', dest='log', required=False)
-
     parser.add_argument('-V', '--verbose', dest='verbose', action='store_true',
                         help='show the serial output')
```
svanscho: Not really part of the PR but might be good to catch any IO errors in opening the file and falling back to stout?

src/modules_bootloader.py
```diff
@@ -322,12 +345,16 @@ def main():
     else:
         logger = lambda msg: sys.stdout.write('%s\n' % msg)
 
-    success = bootload_modules(args.type, args.file, args.verbose, logger)
+    # The type argument is lowercase for backwards compatibility reasons. However, all subsequent calls need the correct type
+    module_type = args.type.upper()
+
+    update_success = bootload_modules(module_type, args.file, args.verbose, logger)
 
     if log_file is not None:
```
svanscho: perhaps wrap previous calls in a try and use a finally here to close file?

svanscho:
Made a few small comments but looks fine as a nice fix and cleanup!

src/modules_bootloader.py
khenderick: I'm going to catch an error and print a user-friendly error, but I don't want to fallback to stdout, as the user might expect logs going into that file.

svanscho:
Looks good to me! 👍

PR #134, opened 2019-02-26
quick fix for entering authorized mode after 5.X seconds regardless of current time

Zoolixx:
This should fix the issue of entering authorized mode immediately after a powercycle.

### DIFF
--- a/src/led_service.py
+++ b/src/led_service.py
@@ -55,6 +55,7 @@ class LedController(object):
         self._input_button = input_button
         self._input_button_pressed_since = None
         self._input_button_released = True
+        self.ticks = 0
 
         self._network_enabled = False
         self._network_activity = False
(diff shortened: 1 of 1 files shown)


src/led_service.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, i2c_device, i2c_address, input_button):
         self._input_button = input_button
         self._input_button_pressed_since = None
         self._input_button_released = True
+        self.ticks = 0
```
khenderick: Make it a private variable `self._ticks`.

src/led_service.py
```diff
@@ -219,13 +220,15 @@ def check_button(self):
                     self._authorized_mode = False
             else:
                 if button_pressed:
+                    self.ticks += 0.250
                     self._input_button_released = False
                     if self._input_button_pressed_since is None:
                         self._input_button_pressed_since = time.time()
-                    if time.time() > self._input_button_pressed_since + 5:
+                    if self.ticks >= 4.500:  # After 4.500 seconds + time to execute the code it should be pressed between 5.8 and 6.5 seconds.
```
khenderick: Maybe just use `0.25` and `4.5`, no need for the rounding to 3 digits after the comma in the code or comments.

PR #135, opened 2019-02-27
Added tests + integration with Jenkins

Zoolixx:


Zoolixx:
@openmotics-ci test gateway=feature/jenkins frontend=feature/FE-tests-preparation

testing/cicd/io_test.py
```diff
@@ -0,0 +1,380 @@
...
+        response_json = self.tools.api_testee('module_discover_start', 'some_token', expected_failure=True)
+        self.assertEqual(response_json, 'invalid_token')
+
+        response_json = self.tools.api_testee('module_discover_status', 'some_token', expected_failure=True)
+        self.assertEqual(response_json, 'invalid_token')
+
+        response_json = self.tools.api_testee('module_discover_stop', 'some_token', expected_failure=True)
+        self.assertEqual(response_json, 'invalid_token')
+
+    @unittest.skip('Currently factory reset is not working properly')
```
khenderick: No need to skip the tests when we don't include it in the test sets, right?

testing/cicd/thermostats_test.py
```diff
@@ -0,0 +1,195 @@
...
+            else:
+                time.sleep(0.3)
+
+        self.assertTrue(response_json.get('automatic', False) is True and response_json.get('setpoint', 99) == 0 and response_json.get('status')[0].get('csetp') == 9, "Should return a thermostat status according to the thermostat auto config that has been set after resetting the master. Got: {0}".format(response_json))
+
+        self.webinterface.set_output(id=self.TESTEE_POWER, is_on=False)
+        time.sleep(0.5)
+        self.webinterface.set_output(id=self.TESTEE_POWER, is_on=True)
+
+        start = time.time()
```
khenderick: If I'm correct this block is basically a wait for the testee to come back online. If so, let's wrap it in a function somewhere so we can basically do something like `something.wait_for_online_testee(timeout=123)`

Zoolixx:
🚀

Zoolixx:
@openmotics-ci test gateway=feature/jenkins frontend=feature/FE-tests-preparation

Zoolixx:
@openmotics-ci test gateway=feature/jenkins frontend=feature/FE-tests-preparation

Zoolixx:
@openmotics-ci test gateway=feature/jenkins frontend=feature/FE-tests-preparation

Zoolixx:
@openmotics-ci  test gateway=feature/jenkins frontend=feature/FE-tests-preparation

testing/cicd/tests/websocket_test.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class WebsocketTest(unittest.TestCase):
+    """
+    The WebsocketTest is a test case for websocket.
+    """
+    webinterface = None
+    tools = None
+    token = ''
+    DATA = None
```
khenderick: I think you can remove this one

testing/cicd/tests/websocket_test.py
```diff
@@ -0,0 +1,153 @@
...
+    @exception_handler
+    def test_websocket_output_change(self):
+        """ Testing the websocket on the Testee for output_change event. """
+        callback_data = {'data': None}
+
+        def _callback(data):
+            """ _callback will set the variable DATA when a message is received. """
+            callback_data['data'] = data
+
+        token = requests.get('https://{0}/login'.format(self.tools.testee_ip),
```
khenderick: If you have a `self.token` available, there's no need to login again here. The token for websockets and normal calls is identical.

testing/cicd/tests/websocket_test.py
```diff
@@ -0,0 +1,153 @@
...
+        self.assertTrue(bool(callback_data['data']), 'Websocket returned an empty response!')
+        self.assertTrue(callback_data['data']['data']['status']['on'], 'Should contain the status of the output. Got: {0}'.format(callback_data['data']))
+        self.assertEqual(callback_data['data']['data']['id'], 3, 'Should contain the correct triggered ID. Got: {0}'.format(callback_data['data']))
+        self.assertEqual(callback_data['data']['type'], 'OUTPUT_CHANGE', 'Should contain the correct event type. Got: {0}'.format(callback_data['data']))
+
+        self.tools.clicker_releaser(3, token, False)
+        self.tools.check_if_event_is_captured(3, 0)
+        time.sleep(0.5)
+
+        self.assertTrue(bool(callback_data['data']), 'Websocket returned an empty response!')
```
khenderick: You should make sure that in such case you re-set the data to `None`. Also, it might make more sense to use `assertIsNotNone` or something like that.

testing/cicd/tests/websocket_test.py
```diff
@@ -0,0 +1,153 @@
...
+
+        token = requests.get('https://{0}/login'.format(self.tools.testee_ip),
+                             params={'username': 'openmotics',
+                                     'password': '123456'},
+                             verify=False).json()['token']
+        socket = PassthroughClient('wss://{0}/ws_events'.format(self.tools.testee_ip),
+                                   protocols=['authorization.bearer.{0}'.format(base64.b64encode(token.encode('ascii')).decode('utf-8').replace('=', ''))], event_name='OUTPUT_CHANGE', callback=_callback)
+        socket.connect()
+
+        self.tools.clicker_releaser(3, token, True)
```
khenderick: As a more general remark; what's the configuration of this output. At this point we're not sure that the output will behave as expected. It might be controlled by a thermostat, or a previous test might have left a timeout that will immediately turn off the output again. For test stability I think it makes sense to configure this output (and input below) to a well defined state and make sure it's not in use by any thermostat.

testing/cicd/tests/websocket_test.py
```diff
@@ -0,0 +1,153 @@
...
+            self.tools.discovery_success = self.tools.assert_discovered(self.token, self.webinterface)
+            if not self.tools.discovery_success:
+                LOGGER.error('Skipped: %s due to discovery failure.', self.id())
+                self.skipTest('Failed to discover modules.')
+        LOGGER.info('Running: %s', self.id())
+
+    @exception_handler
+    def test_websocket_output_change(self):
+        """ Testing the websocket on the Testee for output_change event. """
+        callback_data = {'data': None}
```
khenderick: On a second though, I think it might be better to change from `None` to a list `[]` here. Then the callback can just append the data and the test can just always load the last entry (using `[-1]`). Instead of testing for `None` you validate a length larger than X.

The benefit would be that if data is not what is expected, the debug output will yield more useful information.

testing/cicd/actions_test.py
```diff
@@ -0,0 +1,1305 @@
...
+                         'Expecting the response to be invalid_token. Got: {0}'.format(response_json))
+
+    @exception_handler
+    def test_execute_group_action(self):
+        """ Testing the execution of a group action that toggles output 0. """
+        self._set_group_actions_config()
+
+        input_number = randint(0, 7)
+        group_action_number = randint(0, 159)
+        config = {'name': 'input_ex',
```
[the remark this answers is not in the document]
khenderick: @Zoolixx, This one is still open

testing/cicd/actions_test.py
```diff
@@ -0,0 +1,1305 @@
...
+                                              expected_failure=True)
+        self.assertEqual(response_json.get('msg'),
+                         'group_action_id not in [0, 160]: {0}'.format(invalid_group_action_id),
+                         'Should return an error message when calling do group action API with an invalid group action ID. Got: {0}'.format(
+                             response_json))
+
+    @exception_handler
+    def test_set_group_action_configuration(self):
+        """ Testing the setting up of one group action. """
+        time.sleep(0.5)
```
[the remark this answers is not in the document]
khenderick: @Zoolixx, this one is still open.

testing/cicd/actions_test.py
```diff
@@ -0,0 +1,1305 @@
...
+            url_params = urllib.urlencode({'group_action_id': one.get('id')})
+            self.tools.api_testee('do_group_action?{0}'.format(url_params), self.token)
+        pressed_inputs = json.loads(self.webinterface.get_last_inputs()).get('inputs')
+        self.assertIsNotNone(pressed_inputs)
+        self.assertTrue(pressed_inputs)
+
+    @exception_handler
+    def test_do_group_action_on_off_authorization(self):
+        """ Testing do_group_action API auth verification. """
+        response_json = self.tools.api_testee('get_group_action_configurations', 'some_token', expected_failure=True)
```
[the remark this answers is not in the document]
khenderick: @Zoolixx, this is still open.

testing/cicd/actions_test.py
```diff
@@ -0,0 +1,1305 @@
...
+                             response_json))
+
+        response_json = json.loads(self.webinterface.get_output_status())
+
+        status_list = response_json.get('status', [])
+        self.assertTrue(bool(status_list), 'Should contain the list of output statuses. Got: {0}'.format(status_list))
+        if status_list[8].get('status') == 0:
+            json.loads(self.webinterface.set_output(id=self.TESTEE_POWER, is_on=True))
+        else:
+            json.loads(self.webinterface.set_output(id=self.TESTEE_POWER, is_on=False))
```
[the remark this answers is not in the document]
khenderick: @Zoolixx, this one is still open.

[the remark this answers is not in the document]
Zoolixx: A class wrapper isn't currently necessary since there's only the need to call `json.loads(self.webinterface.foo(...))` for 3 times in all test suites combined.

PR #136, opened 2019-03-07
various code cleanups/optimizations respecting PEP8 and OM code guidelines

Zoolixx:


src/master/thermostats.py
```diff
@@ -57,7 +57,7 @@ def full_update(self, thermostats):
         """
         with self._merge_lock:
             new_status = {t['id']: t for t in thermostats['status']}
-            if len(self._thermostats) == 0:
+            if not bool(self._thermostats):
```
khenderick: No need to convert to bool I think.

Especially related to consistency; either we try to do more detailed checks such as `data is not None and len(data) > 0`, or we do `if data`, but currently it feels like different changes have been made.

src/vpn_service.py
```diff
@@ -164,22 +164,21 @@ def do_call(self, uri):
...
-            return None
+            return
 
     def get_real_time_power(self):
         """ Get the real time power measurements. """
         data = self.do_call("get_realtime_power?token=None")
-        if data is None or data['success'] is False:
-            return None
-        else:
+        if data is not None or data['success']:
```
khenderick: This will raise an exception if data is None.

src/master/master_command.py
```diff
@@ -72,7 +73,8 @@ def create_input(self, cid, fields=None):
 
         return start + encoded_fields + "\r\n"
 
-    def __calc_crc(self, encoded_string):
+    @staticmethod
+    def __calc_crc(encoded_string):
```
khenderick: You convert it to a static method, but I don't see the changes where it's called as a static method (e.g. `MasterCommandSpec.__calc_crc(...)` iso `self.__calc_crc(...)`). You should not call static methods with `self...`.

Zoolixx:
@openmotics-ci test gateway=feature/code-cleanup frontend=feature/FE-tests-preparation

testing/cicd/tests/rooms_test.py
```diff
@@ -138,14 +138,13 @@ def test_set_all_lights_off_force_checked(self):
     @exception_handler
     def test_set_all_lights_floor_off(self):
         """ Testing turning all lights off for a specific floor number. """
-        self._set_room_floor_configuration(
-            room_number=self.ROOM_NUMBER)  # Setting up configuration first. Room: 5, Floor 3
-
+        self._set_room_floor_configuration(room_number=self.ROOM_NUMBER)  # Setting up configuration first. Room: 5, Floor 3
+        token = self.tools.get_new_token('openmotics', '123456')
```
khenderick: Don't use the hard-coded credentials.

I think we should solve issues like this by having the `tools` take care of the authentication. The `api_testee` call could just have an `authorize` parameters set to `True` or `False`, and it could handle an invalid token all by itself by logging in again and re-executing the call. For invalid token tests, you could explicitly pass in an invalid token that disables the auto-login.

Zoolixx: This was only for debugging purposes, new fixes are added here and for Testrunner's toolbox.

src/master/master_command.py
```diff
@@ -392,27 +398,30 @@ def get_min_decode_bytes(self):
...
+
 class SvtFieldType(object):
     """ The System Value Type is one byte. This types encodes and decodes into
     a float (degrees Celsius).
     """
     def __init__(self):
         pass
 
-    def encode(self, field_value):
+    @staticmethod
```
khenderick: Can you change these to class methods like you did for the `DimmerFieldType.encode`? I think the only one that should be static is the `__check_crc`

testing/cicd/tests/websocket_test.py
```diff
@@ -162,6 +152,22 @@ def _set_default_input_configuration(self):
             params = {'config': json.dumps(input_configuration)}
             self.tools.api_testee(api='set_input_configuration', params=params, token=token)
 
+    @staticmethod
+    def _look_for_ws_event(callback_data, expected_id, expected_event, expected_status=None, timeout=3):
+        start = time.time()
+
+        while time.time() - start < timeout:
+            for i in xrange(len(callback_data['data'])):
```
khenderick: Use `for entry in callback_data['data']:` and then you can use `entry...` below instead of `callback_data['data'][i]...`.

PR #140, opened 2019-03-20
safe testing websockets

Zoolixx:


Zoolixx:
@openmotics-ci test gateway=bugfix/eventsocket-metadata frontend=develop

testing/cicd/tests/websocket_test.py
```diff
@@ -107,12 +117,22 @@ def _callback(data):
         self.webinterface.set_output(id=4, is_on=False)
 
         self.assertTrue(len(callback_data['data']) == 2, 'Websocket returned an empty response!')
-        self.assertEqual(callback_data['data'][-1]['data']['id'], 4, 'Should contain the correct triggered ID. Got: {0}'.format(callback_data['data']))
-        self.assertEqual(callback_data['data'][-1]['type'], 'INPUT_TRIGGER', 'Should contain the correct event type. Got: {0}'.format(callback_data['data']))
 
-        callback_data.update({'data': []})
+        if len(callback_data['data']) != 2:
```
khenderick: Shouldn't `_look_for_ws_event` be used here?

Zoolixx: It would be better, I agree, but since we can't differentiate two `INPUT_TRIGGER` events, we have to check that there's indeed 2 different ones.
Using `_look_for_ws_event(callback_data, expected_id=4, expected_event='INPUT_TRIGGER')` twice can return true where in reality there was only one `INPUT_TRIGGER` event

khenderick: Then I suggest we add some more arguments to wait for the correct one, since here it's back to "let's assume the 3rd event received is the correct one".

PR #141, opened 2019-03-26
Added test logs to the testee

Zoolixx:


Zoolixx:
@openmotics-ci test gateway=feature/testee-logs-for-jenkins frontend=develop

testing/cicd/tests/thermostats_test.py
```diff
@@ -35,13 +34,6 @@ class ThermostatsTest(unittest.TestCase):
...
-    @classmethod
-    def setUpClass(cls):
-        if not cls.tools.healthy_status:
-            raise unittest.SkipTest('The Testee is showing an unhealthy status. All tests are skipped.')
-        if not cls.tools.initialisation_success:
-            raise unittest.SkipTest('Unable to initialise the Testee. All tests are skipped.')
-
     def setUp(self):
         self.tools.configure_thermostat(0, self.NIGHT_TEMP_INIT, self.DAY_BLOCK1_INIT, self.DAY_BLOCK2_INIT)  # Configuring thermostat 0
         self.token = self.tools.get_new_token(self.tools.username, self.tools.password)
```
khenderick: I think this can be removed as well? 

If you want to execute additional code (e.g. the `configure_thermostat`), you can still implement a `def setUp(self):`that calls its super, and then execute the extra logic before of after the super has been called.

Something like this should work:

```
def setUp(self):
    super(ThermostatsTest, self).setUp()
    self.tools.configure_thermostat(...)  # ...
```

khenderick: On a side note, I think the current logic was possibly incorrect. If discovery would not have been successful, the `configure_thermostats` was still executed and would probably fail. So I think it should be called after the super.

PR #146, opened 2019-04-23
Add output states in heartbeat data

khenderick:


khenderick:
@openmotics-ci test gateway=bugfix/output-state-in-heartbeat frontend=develop

src/gateway/webservice.py
```diff
@@ -2150,7 +2150,7 @@ def get_version(self):
         :rtype: dict
         """
         return {'version': self._gateway_api.get_main_version(),
-                'gateway': '2.10.1'}
```
svanscho: Can we put all version number strings in one place? Easier to manage version bumps in the future.

khenderick: I agree that eventually we should move this version number to some file that can be generated by e.g. Jenkins and that can be shipped with the code. But I would prefer to do such a change in a separate PR.

svanscho: Agreed.

PR #148, opened 2019-04-25
Restored pushing thermostat data over the heartbeat

khenderick:


src/vpn_service.py
```diff
@@ -189,6 +189,12 @@ def get_pulse_counter_diff(self):
             return ret
         return
 
+    @staticmethod
+    def __counter_diff(current, previous):
+        """ Calculate the diff between two counter values. """
+        diff = current - previous
+        return diff if diff >= 0 else 65536 - previous + current
```
svanscho: else 65536 + diff ?

src/vpn_service.py
```diff
@@ -200,11 +206,22 @@ def get_enabled_outputs(self):
...
+        if data is None or data['success'] is False:
+            return None
+        ret = {'thermostats_on': data['thermostats_on'],
+               'automatic': data['automatic'],
+               'cooling': data['cooling']}
+        thermostats = []
+        for thermostat in data['status']:
+            to_add = {}
+            for field in ['id', 'act', 'csetp', 'mode', 'output0', 'output1', 'outside', 'airco']:
+                to_add[field] = thermostat[field]
```
svanscho: to_add[field] = thermosta.get(field, None) ?

svanscho: As agreed, let's keep changes to a minimum to be as backwards compatible as possible.

src/vpn_service.py
svanscho: Ok for now.

khenderick: Good point, but I don't want to change unaffected code (it just has been moved causing it to show up as a difference). All new flows use metrics, so it's obsolete code pending deletion anyway.

src/vpn_service.py
khenderick: We're quite sure these fields will be in there, and I'd rather have a key error here, than sending unexpected data to the cloud in this case.

PR #150, opened 2019-04-29
Filter values from non-dimmer events

svanscho:
Filtering value from output events for non-dimmer objects.


testing/unittests/gateway_tests/observer_tests.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def setUp(self):
+        pass
+
+    def test_value_formatting_output(self):
+        master_mock = Mock()
+        dbus_mock = Mock()
+
+        def callback_test_value_present(event):
+            self.assertTrue('value' in event.data['status'])
```
khenderick: You can use `assertIn`, which gives nicer errors in case there's a failure.

PR #153, opened 2019-05-09
OM-528 improving GW tests stability

Zoolixx:
Small changes mainly in the discovery test to handle the case of many discovered slave modules


Zoolixx:
@openmotics-ci test gateway=bugfix/improving-integration-gw-tests frontend=develop

PR #160, opened 2019-06-18
OM-590 Shutter lock status handling with version dependency

batmahdi:


src/master/master_api.py
```diff
@@ -195,8 +195,13 @@ def read_input():
                               Field.str('input_name', 8), Field.crc(), Field.lit('\r\n')])
 
 
-def shutter_status():
+def shutter_status(master_version):
     """ Read the status of a shutter module. """
+    parsed_current_version = tuple([int(x) for x in master_version.split('s.')])
```
khenderick: Why split on `s.` instead of `.`? If the version format of the version is something different than e.g. `1.2.3`, it's a good idea to add a comment with an example.

src/master/master_api.py
```diff
@@ -195,8 +195,13 @@ def read_input():
...
+    if parsed_current_version >= tuple("3.143.78"):
```
khenderick: Assuming `parsed_current_version` would be something like `(1, 2, 3)`, you can't compare it to `tuple("1.2.3")` right, as the latter yields `('1', '.', '2', '.', '3')`, which won't give you the expected result I guess.

src/gateway/observer.py
```diff
@@ -114,10 +114,11 @@ def __init__(self, master_communicator, dbus_service):
 
         self._thread = Thread(target=self._monitor)
         self._thread.daemon = True
+        self._master_shutter_status_registered = False
```
khenderick: Would be a little bit nicer to move these two up a bit, right under `self._shutters_last_updated = 0`, together with the other variable initialisations. Here is basically a block where threads are created.

PR #161, opened 2019-06-19
Added VPN fix for when ingesting routes

svanscho:
@khenderick fix for VPN led

### DIFF
--- a/src/vpn_service.py
+++ b/src/vpn_service.py
@@ -92,11 +92,18 @@ class VpnController(object):
     def _vpn_connected(self):
         """ Checks if the VPN tunnel is connected """
         try:
-            route = subprocess.check_output('ip r | grep tun | grep via || true', shell=True).strip()
-            if not route:
-                return False
-            vpn_server = route.split(' ')[0]
-            self.vpn_connected = VPNService.ping(vpn_server, verbose=False)
+            routes = subprocess.check_output('ip r | grep tun | grep via || true', shell=True).strip()
+            #example output:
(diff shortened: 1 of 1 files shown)


src/vpn_service.py
```diff
@@ -92,11 +92,18 @@ def check_vpn():
...
+            #example output:
+            # 10.0.0.0/24 via 10.37.0.5 dev tun0\n
+            # 10.37.0.1 via 10.37.0.5 dev tun0
+            result = False
+            if routes:
+                vpn_servers = [route.split(' ')[0] for route in routes.split('\n') if '/' not in route]
+                for vpn_server in vpn_servers:
+                    if VPNService.ping(vpn_server, verbose=False):
+                        result = True
+                        continue
```
khenderick: I think that should be a `break`

src/vpn_service.py
```diff
@@ -92,11 +92,18 @@ def check_vpn():
     def _vpn_connected(self):
         """ Checks if the VPN tunnel is connected """
         try:
-            route = subprocess.check_output('ip r | grep tun | grep via || true', shell=True).strip()
-            if not route:
-                return False
-            vpn_server = route.split(' ')[0]
-            self.vpn_connected = VPNService.ping(vpn_server, verbose=False)
+            routes = subprocess.check_output('ip r | grep tun | grep via || true', shell=True).strip()
+            #example output:
```
khenderick: Can you add a space between `#` and `example`? Now this gives a weak warning.

PR #162, opened 2019-07-10
Import eggs in the plugin runtime to make sure requests is available

khenderick:


src/plugins/runner.py
```diff
@@ -307,6 +307,8 @@ def error_score(self):
             return score
 
     def get_queue_length(self):
+        if self._async_command_queue is None:
+            return 0
         return self._async_command_queue.qsize()
```
svanscho: Add small unit test

PR #166, opened 2019-08-06
Shutter positions

khenderick:


src/master/eeprom_models.py
```diff
@@ -146,6 +146,7 @@ class ShutterConfiguration(EepromModel):
     group_1 = EepromByte(lambda mid: (63, (mid * 2) + 0))
     group_2 = EepromByte(lambda mid: (63, (mid * 2) + 1))
     room = EextByte()
+    steps = EextWord()
```
svanscho: compatibility with cloud and portal

khenderick: Checked and the cloud is forward compatible.

svanscho:
Looks good, thanks for changes.

PR #171, opened 2019-08-19
Make sure calibration data is kept during an update

khenderick:


src/power_bootloader.py
```diff
@@ -193,54 +235,57 @@ def main():
...
-                                           verbose=args.verbose)
+    power_communicator = PowerCommunicator(power_serial, None, time_keeper_period=0, verbose=args.verbose)
     power_communicator.start()
 
+    def _bootload(_module, _module_address):
+        try:
+            if args.old and _module['version'] == POWER_API_8_PORTS:
+                bootload_8(_module_address, args.file, power_communicator)
+            elif not args.old and _module['version'] == POWER_API_12_PORTS:
+                bootload_12(_module_address, args.file, power_communicator)
```
svanscho: add args.old args.file as method params

PR #172, opened 2019-08-21
Don't block on unavailable updates

khenderick:


src/power_bootloader.py
```diff
@@ -258,30 +258,27 @@ def _bootload(_module, _module_address, filename, is_power_module):
                 bootload_8(_module_address, filename, power_communicator)
             elif not is_power_module and _module['version'] == POWER_API_12_PORTS:
                 bootload_12(_module_address, filename, power_communicator)
-            return True
+        except CommunicationTimedOutException:
+            logger.warning('E{0} - Module unavailable. Skipping...'.format(address))
         except Exception:
             logger.exception('E{0} - Unexpected exception during bootload'.format(address))
```
svanscho: Add Skipping... in logs

src/power_bootloader.py
```diff
@@ -258,30 +258,27 @@ def _bootload(_module, _module_address, filename, is_power_module):
                 bootload_8(_module_address, filename, power_communicator)
             elif not is_power_module and _module['version'] == POWER_API_12_PORTS:
                 bootload_12(_module_address, filename, power_communicator)
-            return True
```
svanscho: Add log line that update went well

khenderick: That's already logged in the `bootload_8` and `bootload_12` methods

svanscho: Excellent

PR #174, opened 2019-08-22
Dependency injection

svanscho:
@khenderick this includes the DI framework and some refactoring of classes. We would still need some more work later on to remove the circular dependencies.


src/gateway/scheduling.py
```diff
@@ -139,6 +145,7 @@ def __init__(self, db_filename, lock, gateway_api):
         self._stop = False
         self._processor = None
         self._semaphore = None
+        self._web_interface = None
```
khenderick: Double

src/gateway/shutters.py
```diff
@@ -57,6 +60,9 @@ class State(object):
                            State.GOING_DOWN: Direction.DOWN,
                            State.STOPPED: Direction.STOP}
 
+    @provides('shutter_controller')
+    @scope(SingletonScope)
+    @inject(master_communicator='master_communicator')
     def __init__(self, master_communicator, verbose=False):
```
khenderick: Just a question, the DI can handle the `verbose` not being injected?

svanscho: it's basically like not providing the parameter, so it will use the defaults

src/gateway/config.py
```diff
@@ -20,6 +20,9 @@
 import sqlite3
 import logging
 from random import randint
+
```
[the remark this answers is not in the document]
svanscho: OK

PR #176, opened 2019-08-26
Don't send events to the Cloud that are not relevant

khenderick:


src/cloud/client.py
```diff
@@ -44,20 +45,26 @@ def send_event(self, event):
         if self.api_version != 0:
             raise NotImplementedError('Sending events is not supported on this api version')
 
+        if event.type in [Event.Types.INPUT_TRIGGER,
```
svanscho: Move this upstream

PR #177, opened 2019-08-26
Release 2.12.0

khenderick:


requirements.txt
```diff
@@ -1,5 +1,10 @@
...
 cherrypy == 3.2.2
-requests == 1.2.0
+requests == 2.5.3
 intelhex == 1.5
 xmlrunner == 1.7.7
+ws4py == 0.5.1
+msgpack == 0.6.1
+decorator == 4.4.0
+simplejson == 3.16.0
+serial
```
svanscho: pyserial?

khenderick: Yep, fixed in one of the pending PR. But no need to urge that PR as this file isn't used in updates.

PR #179, opened 2019-08-28
Send events async

khenderick:


src/cloud/event_sender.py
```diff
@@ -0,0 +1,77 @@
...
+                logger.error(ex)
+                time.sleep(1)
+            except Exception:
+                logger.exception('Unexpected error when sending events')
+                time.sleep(1)
+
+    def _send_events(self):
+        events = []
+        try:
+            events.append(self._queue.pop())
```
svanscho: Seems only one event will be send at a time. Perhaps also use some batching? Like max 100 events at a time?

PR #181, opened 2019-08-30
Master to develop

khenderick:


src/cloud/cloud_api_client.py
```diff
@@ -15,9 +15,6 @@
 """
 Cloud API Client
 """
-
-from platform_utils import System
-System.import_eggs()
```
svanscho: I think we need the updated request egg. It might have been imported earlier, but we cannot blindly rely on that.

src/cloud/event_sender.py
```diff
@@ -0,0 +1,82 @@
...
+            except APIException as ex:
+                logger.error(ex)
+                time.sleep(1)
+            except Exception:
+                logger.exception('Unexpected error when sending events')
+                time.sleep(1)
+
+    def _send_events(self):
+        events = []
+        while len(events) < 25:
```
svanscho: 👍

src/cloud/cloud_api_client.py
svanscho: After discussion - agree to drop this requirement

PR #196, opened 2019-11-12
CPU improvements

khenderick:


khenderick:
Before merging in we need to add the uJSON egg

src/platform_utils.py
```diff
@@ -198,13 +198,24 @@ def handle_socket_exception(connection, exception, logger):
 
     @staticmethod
     def import_eggs():
-        current_file_path = os.path.dirname(os.path.abspath(__file__))
         operating_system = System._get_operating_system()
-        blacklisted_eggs = {'debian': ['requests-2.5.3-py2.7.egg', 'angstrom-psutil-5.6.3-py2.7-linux-armv7l.egg'],
-                            'angstrom': ['debian-psutil-5.6.3-py2.7-linux-armv7l.egg']}.get(operating_system['ID'], [])
+        blacklisted_eggs = {
+            'debian': ['requests-2.5.3-py2.7.egg',
```
svanscho: why not change this to e.g. angstrom-requests-2.5.3 etc?
then we can simplify the blacklisting logic too?

src/platform_utils.py
```diff
@@ -198,13 +198,24 @@ def handle_socket_exception(connection, exception, logger):
...
+                       'angstrom-psutil-5.6.3-py2.7-linux-armv7l.egg',
+                       'angstrom-ujson-1.35-py2.7-linux-armv7l.egg'],
+            'angstrom': ['debian-psutil-5.6.3-py2.7-linux-armv7l.egg',
+                         'debian-ujson-1.35-py2.7-linux-armv7l.egg']
+        }.get(operating_system['ID'])
+
+        def _is_blacklisted(_egg):
+            if blacklisted_eggs is not None:
```
svanscho: same comment as above

src/platform_utils.py
khenderick: Good remark, rewritten.

src/platform_utils.py
```diff
@@ -198,16 +199,17 @@ def handle_socket_exception(connection, exception, logger):
...
-                    resource_stream(Requirement.parse('requests'), 'requests/cacert.pem')
-                    os.environ['REQUESTS_CA_BUNDLE'] = resource_filename(Requirement.parse('requests'), 'requests/cacert.pem')
+
+        eggs = (glob.glob('{0}/eggs/*.egg'.format(current_file_path)) +
+                glob.glob('{0}/eggs/{1}/*.egg'.format(current_file_path, operating_system)))
+        for egg in eggs:
+            sys.path.insert(0, egg)
+
+        # Patching where/if required
+        if operating_system == 'angstrom':
```
svanscho: 🥇

src/power_bootloader.py
```diff
@@ -273,7 +273,7 @@ def _bootload(_module, _module_address, filename, is_power_module):
                 _bootload(module, address, args.file, is_power_module=args.old)
         else:
             address = args.address
-            modules = [module for module in power_modules if module['address'] == address]
+            modules = [module for module in power_modules.values() if module['address'] == address]
```
svanscho: bug?

testing/unittests/plugins_tests/base_tests.py
```diff
@@ -262,6 +266,91 @@ class Test(OMPluginBase):
         self.assertEqual(result, 'Plugin successfully installed')
         self.assertEqual([r.name for r in controller.get_plugins()], ['Test'])
 
+    def test_plugin_metric_reference(self):
+        """ Validates whether two plugins won't get the same metric instance """
+        controller = None
+        try:
+            p1_md5, p1_data = PluginControllerTest._create_plugin_package('P1', """
+from plugins.base import *
```
svanscho: how about reading this from an external file for cleanliness?

svanscho:
Approved, but perhaps a few small improvements can still be included?

src/power_bootloader.py
khenderick: Yes, noticed it during testing

src/platform_utils.py
svanscho: perhaps also include a .lower() just to be sure

testing/unittests/plugins_tests/base_tests.py
khenderick: I understand your comment. It's been an inner conflict for quite a while. It would for sure be easier to catch "bugs" in these fake plugins at the cost of losing the quick overview of what these minimal fake plugins are doing. 

So far, the benefit of overview has won, I like to see the "context" in which a test will run in the test instead of in separate files. But I guess we can always restructure if it would get out of hand.

src/platform_utils.py
khenderick: It's very unlikely it would change as it's in our OS base image, but I can for sure lowercase that.

PR #217, opened 2020-01-14
Stability fixes

khenderick:


src/gateway/metrics_controller.py
```diff
@@ -496,16 +495,24 @@ def _collect_openmotics(self):
...
-                    rate_key = '{0}.{1}'.format(metric['source'].lower(), metric['type'].lower())
-                    if rate_key not in self.outbound_rates:
-                        self.outbound_rates[rate_key] = 0
-                    self.outbound_rates[rate_key] += delivery_count
-                    self.outbound_rates['total'] += delivery_count
-            except IndexError:
-                time.sleep(0.1)
+                metrics = []
+                try:
+                    while True:
```
svanscho: suggestion: `while True and len(metrics) < 250:`

what do you think?

src/plugin_runtime/runtime.py
```diff
@@ -238,9 +238,10 @@ def _handle_collect_metrics(self, name):
             IO._log_exception('collect metrics', exception)
         return {'metrics': metrics}
 
-    def _handle_distribute_metric(self, name, metric):
+    def _handle_distribute_metrics(self, name, metrics):
         receive = getattr(self._plugin, name)
-        IO._with_catch('distribute metric', receive, [metric])
+        for metric in metrics:
+            IO._with_catch('distribute metric', receive, [metric])
```
svanscho: does this mean they still get serialized individually over the IPC ?

src/vpn_service.py
```diff
@@ -511,10 +511,9 @@ def _check_vpn(self):
                     self._clean_debug_dumps()
 
                 if self._iterations > 20 and self._cloud.get_last_connect() < time.time() - REBOOT_TIMEOUT:
-                    # The cloud is not responding for a while.
-                    if not VPNService.ping('cloud.openmotics.com') and not VPNService.ping('8.8.8.8') and not VPNService.ping(VPNService._get_gateway()):
-                        # Perhaps the BeagleBone network stack is hanging, reboot the gateway
-                        # to reset the BeagleBone.
+                    # We can't connect for over `REBOOT_TIMEOUT` seconds and we tried for at least 20 times.
```
svanscho: is this the fix that got 'lost' ?

src/plugin_runtime/runtime.py
khenderick: No, at this point we're already on the plugin's side, after the IPC.

src/vpn_service.py
khenderick: Yes

src/gateway/metrics_controller.py
khenderick: Yep, that would be more elegant.

PR #224, opened 2020-01-22
Move from eggs to wheels

khenderick:


src/post-update.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+OS_DIST=`awk -F= '$1=="ID" { print $2 ;}' /etc/os-release`
```
svanscho: Elegant! Not sure if we still want to store the whl files under the python directory as well. Don't you think this is confusing?

khenderick: I was in doubt about this as well.

Ideally I'd like to have this out of the repo completely and have these dependencies installed/updated by a more global mechanism that fetches them from an online repository we manage. Then it's just a matter of keeping the `requirements.txt` up to date (as we need to do anyway for unit testing purposes).

However, let's do this when we take the next steps on our update processes. For now, including them should make sure that we can keep using our existing update logic as much as possible (where we ship everything in the `src` directory to the `python` folder). So it's a matter of simplicity at this moment.

svanscho: Ok agreed.

PR #227, opened 2020-01-30
Implement sensors on MasterController

khenderick:


src/master_core/core_communicator.py
```diff
@@ -283,82 +283,86 @@ def _read(self):
         footer_length = 1 + 1 + len(CoreCommunicator.END_OF_REPLY)  # 'C' + checksum (1 byte) + \r\n
 
         while not self._stop:
+            try:
```
khenderick: This change is only adding a `try/except`. There's no code changes inside the `try`-block.

src/gateway/hal/master_controller_core.py
```diff
@@ -231,3 +242,77 @@ def shutter_stop(self, shutter_id):
...
                                                                       'extra_parameter': 0})
+
+    # Sensors
+
+    def get_sensor_temperature(self, sensor_id):
+        return self._sensor_states.get(sensor_id, {}).get('TEMPERATURE')
+
+    def get_sensors_temperature(self):
+        temperatures = []
+        for sensor_id in xrange(32):
```
svanscho: Is this a Core+ limitation of 32? If not suggest, to list all sensors and truncate to 32 in gateway_api, 32 is an API spec limitation, not a Core+ limitation.

src/gateway/hal/master_controller_core.py
```diff
@@ -231,3 +242,77 @@ def shutter_stop(self, shutter_id):
...
+            temperatures.append(self.get_sensor_temperature(sensor_id))
+        return temperatures
+
+    def get_sensor_humidity(self, sensor_id):
+        return self._sensor_states.get(sensor_id, {}).get('HUMIDITY')
+
+    def get_sensors_humidity(self):
+        humidities = []
+        for sensor_id in xrange(32):
```
svanscho: same comment

src/gateway/hal/master_controller_core.py
```diff
@@ -231,3 +242,77 @@ def shutter_stop(self, shutter_id):
...
+            humidities.append(self.get_sensor_humidity(sensor_id))
+        return humidities
+
+    def get_sensor_brightness(self, sensor_id):
+        # TODO: This is a lux value and must somehow be converted to legacy percentage
```
svanscho: ticketje?

src/gateway/hal/master_controller_core.py
```diff
@@ -231,3 +242,77 @@ def shutter_stop(self, shutter_id):
...
+        brightness = self._sensor_states.get(sensor_id, {}).get('BRIGHTNESS')
+        if brightness in [None, 65535]:
+            return None
+        return int(float(brightness) / 65535.0 * 100)
+
+    def get_sensors_brightness(self):
+        brightnesses = []
+        for sensor_id in xrange(32):
```
svanscho: same comment

src/master_core/events.py
```diff
@@ -66,16 +72,16 @@ def data(self):
...
+                sensor_type = Event.SensorType.TEMPERATURE
                 sensor_value = self._data[1]
             elif self._action == 1:
-                sensor_type = 'HUMIDITY'
+                sensor_type = Event.SensorType.HUMIDITY
                 sensor_value = self._data[1]
             elif self._action == 2:
-                sensor_type = 'BRIGHTNESS'
+                sensor_type = Event.SensorType.BRIGHTNESS
                 sensor_value = Event._word_decode(self._data[0:2])
```
svanscho: why is this different? Also using a protected method doesn't look so clean.

src/gateway/hal/master_controller_core.py
khenderick: Created under internal reference OM-1055

src/master_core/events.py
khenderick: That line didn't change, I just introduced an enum-style approach.

Why isn't it clean to use a private/protected method? This code is inside the `Event` class, so I just use a private static method to do the decoding of that part of the data.

svanscho: Oh missed that we are in the class here. Looks good then :-)

PR #242, opened 2020-02-25
Ignore invalid debug dumps

khenderick:


src/vpn_service.py
```diff
@@ -418,7 +418,10 @@ def _get_debug_dumps(self):
             timestamp = int(filename.replace('/tmp/debug_', '').replace('.json', ''))
             if timestamp not in self._debug_data:
                 with open(filename, 'r') as debug_file:
-                    self._debug_data[timestamp] = json.load(debug_file)
+                    try:
+                        self._debug_data[timestamp] = json.load(debug_file)
+                    except ValueError:
+                        pass  # Ignore this dump for now
```
LnL7: probably best to at least log this as an error for debugging purposes

khenderick: Fair point. Changed for a warning log statement

PR #252, opened 2020-03-10
revive integration tests

LnL7:


testing/cicd/tests/api_test.py
```diff
@@ -0,0 +1,77 @@
...
+@pytest.mark.smoke
+def test_version(toolbox):
+    data = toolbox.target.get('/get_version')
+    assert 'version' in data
+    assert 'gateway' in data
+
+
+@pytest.fixture
+def set_timezone(request, toolbox):
+    yield
```
khenderick: Set the timezone to UTC before the tests, so that a test always starts with a known state.

testing/cicd/tests/inputs_test.py
```diff
@@ -0,0 +1,111 @@
...
+
+    input_config = json.dumps({'id': input_id, 'action': output_id, 'invert': 255})
+    toolbox.target.get('/set_input_configuration', {'config': input_config})
+
+    # NOTE ensure output status _after_ input configuration, changing
+    # inputs can impact the output status for some reason.
+    output_config = {'timer': 2**16 - 1}
+    toolbox.ensure_output(output_id, not output_status, output_config)
+
+    toolbox.toggle_input(input_id)
```
khenderick: Call it something like "press" as this is more in line with what would actually be tested.

testing/cicd/tests/inputs_test.py
```diff
@@ -0,0 +1,111 @@
...
+    toolbox.assert_output_event(output_id, output_status)
```
khenderick: I think `assert_output` would be a better name. The name "event" might not mean much in this context, it's more something for in the implementation itself.

testing/cicd/tests/inputs_test.py
```diff
@@ -0,0 +1,111 @@
...
+    logger.warning('should use a shorter timeout, waiting for 2m30s')
+    time.sleep(180)
+    toolbox.assert_output_event(output_id, not output_status)
```
khenderick: Instead of just asserting here, I'd rather have something like `assert_output(...)` with maybe an optional `wait` parameter that - if given - does not immediately assert, but waits for max `wait` seconds for the assertion to become true. It can then return the time it took which can then be asserted to be e.g. between 2m27s and 2m32s.

testing/cicd/tests/inputs_test.py
```diff
@@ -0,0 +1,111 @@
...
+
+    actions = ['2', str(group_action_id)]
+    input_config = json.dumps({'id': input_id, 'basic_actions': ','.join(actions), 'action': 240, 'invert': 255})
+    toolbox.target.get('/set_input_configuration', {'config': input_config})
+
+    actions = ['162', str(output_id), '162', str(other_output_id)]  # toggle both outputs
+    config = {'id': group_action_id, 'actions': ','.join(actions)}
+    toolbox.target.get('/set_group_action_configuration', params={'config': json.dumps(config)})
+
+    time.sleep(2)
```
khenderick: Why is this sleep needed?
* If it's because it introduces unexpected behavior in the test, it's likely a bug in the Gateway/Master and we should add a TODO/FIXME and create an internal ticket to solve this
* If not, let's remove it.

testing/cicd/tests/mode_test.py
```diff
@@ -0,0 +1,159 @@
...
+def check_ip_range():
+    # type: () -> bool
+    addresses = ([x.address for x in xs if x.broadcast == '10.91.115.255'] for xs in psutil.net_if_addrs().values())
+    return not sum(addresses, [])
+
+
+@pytest.fixture
+def power_on(request, toolbox):
+    yield
+    toolbox.ensure_power_on()
```
khenderick: Similar as above, let's make sure it's on _before_ we start the tests. Just in case somebody turned it off.

testing/cicd/tests/mode_test.py
```diff
@@ -0,0 +1,159 @@
...
+
+@pytest.fixture
+def authorized_mode(request, toolbox):
+    yield
+    toolbox.wait_authorized_mode()
+
+
+@pytest.mark.slow
+def test_power_cycle(toolbox, power_on):
+    toolbox.power_cycle()
```
khenderick: I'm missing asserts? A correct test would be to power-cycle and then wait for a given amount of time until e.g. the health check works, or we can call an API, or whatever that we can consider "up and running".

testing/cicd/tests/mode_test.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+@pytest.mark.smoke
+def test_module_discover(toolbox, discover_mode):
+    logger.info('start module discovery')
+    toolbox.target.get('/module_discover_start')
+    time.sleep(2)
```
khenderick: Again, I'd rather have this polling based (e.g. call the status every 0.5s) until it's in discovery mode, with a max of e.g. 2 seconds

testing/cicd/tests/mode_test.py
```diff
@@ -0,0 +1,159 @@
...
+
+    data = toolbox.target.get('/module_discover_status')
+    assert data['running'] == True
+    toolbox.discover_input_module()
+    toolbox.discover_output_module()
+    time.sleep(2)
```
khenderick: Poll `get_modules` and wait for max 2s for it to be available.

testing/cicd/tests/mode_test.py
```diff
@@ -0,0 +1,159 @@
...
+    toolbox.wait_authorized_mode()
+
+
+@pytest.mark.slow
+def test_power_cycle(toolbox, power_on):
+    toolbox.power_cycle()
+
+
+@pytest.mark.smoke
+def test_module_discover(toolbox, discover_mode):
```
khenderick: As a general remark for this test; I think this one only makes sense if we remove all modules before starting the discovery. At this moment, this test might pass even though the discovery is broken because the modules were already available.

testing/cicd/tests/mode_test.py
```diff
@@ -0,0 +1,159 @@
...
+    assert 'inputs' in data
+    assert 'I' in data['inputs']
+    assert 'outputs' in data
+    assert 'O' in data['outputs']
+
+
+@pytest.mark.slow
+@pytest.mark.skipif(check_ip_range(), reason='the maintenance ports are not accessible on jenkins')
+def test_maintenance(toolbox):
+    # TODO: ensure maintenance mode is disabled afterwards
```
khenderick: Let's add this.

testing/cicd/tests/mode_test.py
```diff
@@ -0,0 +1,159 @@
...
+    ssl_sock.connect((toolbox.target._host, data['port']))
+
+    data = ''
+    while data != 'OK':
+        data = ssl_sock.recv(1024).strip()
+        logger.debug('received data "{}"'.format(data))
+
+    ssl_sock.send('firmware version\r\n')
+    assert ssl_sock.recv(1024).strip() == 'firmware version'
+    assert ssl_sock.recv(1024).strip() == expected_version
```
khenderick: This feels tricky as you expect a slight delay here between the two lines. Maybe a small state machine might be a better option? Try for a second quickly reading the socket, first waiting for `firmware_version`, then for some version-ish string.

testing/cicd/tests/mode_test.py
```diff
@@ -0,0 +1,159 @@
...
+    toolbox.start_authorized_mode()
+    data = toolbox.target.get('/get_usernames')
+    assert 'openmotics' in data['usernames']
+
+
+@pytest.mark.slow
+def test_factory_reset(toolbox, authorized_mode, discover_mode):
+    logger.info('factory reset')
+    toolbox.target.get('/factory_reset')
+    time.sleep(2)
```
khenderick: I think this test will be updated according to some other internal tickets. Let's disable/skip the test for now.

testing/cicd/tests/outputs_test.py
```diff
@@ -0,0 +1,121 @@
...
+        value = draw(one_of(map(just, toolbox.target_outputs)).filter(lambda x: x not in used_values))
+        used_values.append(value)
+        hypothesis.note('module o#{}'.format(value))
+        return value
+    return f
+
+
+@pytest.mark.smoke
+@hypothesis.given(next_output(), booleans())
+def test_events(toolbox, next_output, output_status):
```
khenderick: Let's call this "test_output" or similar, as the event things is something internally.

testing/cicd/tests/outputs_test.py
```diff
@@ -0,0 +1,121 @@
...
+    output_config = {'type': 0, 'timer': 2**16 - 1}
+    toolbox.ensure_output(output_id, not output_status, output_config)
+
+    toolbox.set_output(output_id, output_status)
+    toolbox.assert_output_event(output_id, output_status)
+
+
+@pytest.mark.smoke
+@hypothesis.given(next_output(), booleans(), booleans())
+def test_status(toolbox, next_output, previous_output_status, output_status):
```
khenderick: Similar, not sure what the difference is between the assert output status and assert output event.

testing/cicd/tests/outputs_test.py
```diff
@@ -0,0 +1,121 @@
...
+@hypothesis.given(next_output(), just(True))
+def test_timers(toolbox, next_output, output_status):
+    output_id = next_output(toolbox)
+    logger.info('output timer o#{}, expect event {} -> {}'.format(output_id, output_status, not output_status))
+    output_config = {'type': 0, 'timer': 10}  # FIXME: event reordering with timer of <2s
+    toolbox.ensure_output(output_id, False, output_config)
+
+    toolbox.set_output(output_id, output_status)
+    toolbox.assert_output_event(output_id, output_status)
+    toolbox.assert_output_event(output_id, not output_status)
```
khenderick: Here we could also use this wait parameter to wait for a maximum of e.g. 12 seconds, and assert it's between 9 and 11 seconds.

testing/cicd/tests/outputs_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+@pytest.mark.smoke
+@hypothesis.given(next_output(), integers(min_value=0, max_value=254), just(True))
+def test_floor_lights(toolbox, next_output, floor_id, output_status):
+    light_id, other_light_id, other_output_id = (next_output(toolbox), next_output(toolbox), next_output(toolbox))
+    logger.info('light o#{} on floor {}, expect event {} -> {}'.format(light_id, floor_id, not output_status, output_status))
+
+    output_config = {'type': 255, 'floor': floor_id, 'timer': 2**16 - 1}
```
khenderick: Please make some constant for these types. It will make the test easier to understand.

testing/cicd/tests/toolbox.py
```diff
@@ -0,0 +1,304 @@
...
+        while since > time.time() - timeout:
+            try:
+                response = requests.get(uri, params=params, headers=headers, **self._default_kwargs)
+                data = response.json()
+                if success and 'success' in data:
+                    assert data['success'], 'content={}'.format(response.content)
+                return data
+            except (AssertionError, ConnectionError, RequestException) as exc:
+                logger.error('request {} failed {}, retrying...'.format(path, exc))
+                time.sleep(16)
```
khenderick: Let's make this as argument, so the retry interval is flexible based on the context of the call. For example, after a power cycle, a slow retry might make sense. But if will not make sense when an output would be changed.

testing/cicd/tests/toolbox.py
```diff
@@ -0,0 +1,304 @@
...
+
+    def __init__(self):
+        # type: () -> None
+        self._observer = None  # type: Optional[Observer]
+        self._target = None  # type: Optional[Client]
+        self._target_inputs = None  # type: Optional[List[int]]
+        self._target_outputs = None  # type: Optional[List[int]]
+
+    @property
+    def observer(self):
```
khenderick: Let's call this `tester`

testing/cicd/tests/toolbox.py
```diff
@@ -0,0 +1,304 @@
...
+        # type: () -> Observer
+        if self._observer is None:
+            observer_auth = os.environ['OPENMOTICS_OBSERVER_AUTH'].split(':')
+            observer_host = os.environ['OPENMOTICS_OBSERVER_HOST']
+            self._observer = Observer(Client(observer_host, auth=observer_auth))
+        return self._observer
+
+    @property
+    def target(self):
```
khenderick: Let's call this `testee` (or `dut` to make a few colleagues happy)

testing/cicd/tests/toolbox.py
```diff
@@ -0,0 +1,304 @@
...
+        self.toggle_input(self.DEBIAN_DISCOVER_OUTPUT)
+
+    def power_cycle(self):
+        # type: () -> None
+        logger.info('start power cycle')
+        self.observer.get('/set_output', {'id': self.DEBIAN_POWER_OUTPUT, 'is_on': False})
+        time.sleep(30)
+        self.ensure_power_on()
+        logger.info('waiting 5m for system to stabilize')
+        time.sleep(300)
```
khenderick: Try/poll to make it faster

testing/cicd/tests/inputs_test.py
LnL7: assert_output_event is already a polling check, but I'll make the expected window explicit for this case

testing/cicd/tests/mode_test.py
LnL7: would make more sense to do that before the entire test run instead of specifically this test case

testing/cicd/tests/mode_test.py
LnL7: That's part of factory reset, but since this is a very quick in comparison I think it's still useful to run the noop case.

testing/cicd/tests/toolbox.py
LnL7: This is mostly to deal with infrastructure related issues. This should only occur in special cases and retrying immediately with a flaky connection most likely won't help.

testing/cicd/tests/mode_test.py
LnL7: done in the toolbox fixture instead

PR #254, opened 2020-03-10
Fixed starting sate of next execution

nastassiaseuruk:


src/gateway/scheduling.py
```diff
@@ -59,26 +59,32 @@ def __init__(self, id, name, start, repeat, duration, end, schedule_type, argume
 
     @property
     def is_due(self):
-        if self.status != 'ACTIVE':
```
khenderick: Why was this removed? This means that completed schedules will still be executed?

src/gateway/scheduling.py
```diff
@@ -59,26 +59,32 @@ def __init__(self, id, name, start, repeat, duration, end, schedule_type, argume
...
             # Single-run schedules should start on their set starting time if not yet executed
             if self.last_executed is not None:
                 return False
             return self.start <= time.time()
         # Repeating
         timezone = pytz.timezone(Schedule.timezone)
-        now = datetime.now(timezone)
-        cron = croniter(self.repeat, now)
-        next_execution = cron.get_next(ret_type=float)
+        start_date = datetime.fromtimestamp(self.start, timezone)
```
khenderick: I don't think this is an optimal solution. Since you always start the cron engine from the start date, it will have to traverse the whole timeline every minute for every schedule. Imagine a 30min repeating schedule started somewhere two years ago. It would have to calculate about 35k crons every minute.

It might be better to use something like the following setup:

```
if self.start < time.time():
    base_date = datetime.fromtimestamp(self.start, timezone)
else:
    base_date = datetime.now(timezone)
cron = croniter(self.repeat, base_date)
next_execution = cron.get_next(ret_type=float)
... # rest of the original code
```

src/gateway/scheduling.py
anastasea: The single call of `is_due` here https://github.com/openmotics/gateway/blob/0a6e463ddad539e57a6fc353c2d26edb6799b6c6/src/gateway/scheduling.py#L227 validates this before entering method

nastassiaseuruk: The single call of `is_due` here https://github.com/openmotics/gateway/blob/0a6e463ddad539e57a6fc353c2d26edb6799b6c6/src/gateway/scheduling.py#L227 validates this already

khenderick: Correct :)

src/gateway/scheduling.py
nastassiaseuruk: Yes, but in this case cron starts getting next from now - what is wrong, for example it is settled to do smth every 3rd day from 12th of March and first date should be 12, then 15th, but if we start from today we get 10th and 13th of March.

PR #268, opened 2020-03-16
rework factory reset

LnL7:


src/gateway/gateway_api.py
```diff
@@ -682,45 +684,18 @@ def restore_full_backup(self, data):
...
+        # type: () -> Dict[str,Any]
+        lock_file = constants.get_init_lockfile()
+        if os.path.isfile(lock_file):
+            return {'success': False, 'factory_reset': 'already_in_progress'}
+        with open(lock_file, 'w') as fd:
+            fd.write('factory_reset')
+
+        def _restart():
+            logger.info('Restarting for factory reset...')
+            subprocess.Popen(['systemctl', 'restart', '--no-block', 'openmotics.service'])
```
khenderick: Create a helper function in `platform_utils.System`  so you can call eg. `System.restart('openmotics')` or similar. The implementation should then figure out whether to use systemd, supervisorctl, ... for that particular service.

testing/cicd/systemd/openmotics-led.service
```diff
@@ -1,6 +1,5 @@
 [Unit]
 Description=OpenMotics LEDs
-After=network.target
```
khenderick: I think this one should also required `openmotics.service`, similar to the `openmotics-vpn` service. Similar for the `WantedBy` and others

testing/cicd/systemd/openmotics.service
```diff
@@ -1,16 +1,14 @@
 [Unit]
-Description=OpenMotics Gateway Service
-After=network.target
+Description=OpenMotics Gateway
 
 [Service]
-User=root
-ExecStart=/usr/bin/python2 openmotics_service.py
+Type=oneshot
+ExecStart=/usr/bin/python2 openmotics_init.py
```
khenderick: If we go for an initialisation service we also need to implement this for the current supervisord, as albeit factory reset is currently not working as expected, it should work in our current (supervisord) setup. If that would be too cumbersome we could - for now - just embed the init part inside the `openmotics_service.py`

LnL7: Created a ticket for supervisord support and the systemd move.

PR #278, opened 2020-03-24
Wheels no test

svanscho:


LnL7:
While mock should be a test dependency it's an explicit dependency of [`peewee_migrate`](https://github.com/klen/peewee_migrate/blob/6c7a311a45d9138532b1126e0652f5eca42b8d51/setup.py#L52), resulting in.

```
ERROR: Could not find a version that satisfies the requirement mock>=2.0.0 (from peewee-migrate==1.1.6) (from versions: none)
ERROR: No matching distribution found for mock>=2.0.0 (from peewee-migrate==1.1.6)
```

khenderick:
I assumed the change was tested.

PR #284, opened 2020-03-26
System info

svanscho:


src/gateway/webservice.py
```diff
@@ -2078,6 +2078,12 @@ def get_version(self):
         return {'version': self._gateway_api.get_main_version(),
                 'gateway': gateway.__version__}
 
+    @openmotics_api(auth=True)
+    def get_system_info(self):
+        return {'model': Hardware.get_board_type(),
+                'operating_system': System.get_operating_system()['ID'],
```
khenderick: Let's not only load the `ID`, but everything?

khenderick:
I think it would be (more) interesting to request this information in the `vpn_service` and send it back home with the heartbeat.

src/gateway/webservice.py
svanscho: I want the format to be a bit more defined, so I added a few more attributes but still in a more structured way instead of returning the whole dict.

svanscho:
@khenderick agree this can **also** be sent in the heartbeat. But it's quite static information, so thought about querying the gateway if the latest info is really needed.

PR #287, opened 2020-03-27
fix mypy errors

LnL7:


src/master/eeprom_controller.py
```diff
@@ -1075,19 +1079,3 @@ def decode(self, value):
 
     def encode(self, value):
         return value
-
-
-class EextBool(EextDataType):
```
LnL7: duplicate class

svanscho:
Nice cleanup.

PR #290, opened 2020-03-31
rework daemon threads

LnL7:


src/gateway/daemon_thread.py
```diff
@@ -0,0 +1,84 @@
...
+        self._name = name
+        self._target = target
+        self._tick = threading.Event()
+        self._stop = threading.Event()
+        self._parent = threading.current_thread()
+        self._thread = threading.Thread(target=self._run, name=name)
+
+    def start(self):
+        # type: () -> None
+        logger.debug('Starting daemon {}'.format(self._name))
```
svanscho, line 44: Is there a reason for the different log levels in starting and stopping?

src/gateway/metrics_controller.py
```diff
@@ -446,95 +454,85 @@ def _collect_plugins(self):
...
-                    log('Metric contains unknown values: {0}'.format(', '.join(unknown_metrics)))
+                    break
+                if not isinstance(metric[key], key_type):
+                    log('Metric key {0} should be of type {1}'.format(key, key_type))
                     metric_ok = False
-                if metric_ok is False:
-                    continue
-                self._put(metric)
-            if not self._stopped:
-                time.sleep(max(0.1, 1 - (time.time() - start)))
```
khenderick, line 492: These sleep for a specific time so that they always (or as far as possible) run on exactly the same interval; e.g. once every 30 seconds, even though sometimes this loop might take 2 seconds, and sometimes 15. The `DaemonThread` should also support that before replacing this.

src/gateway/webservice.py
```diff
@@ -2423,24 +2423,23 @@ def _http_server_logger(msg='', level=20, traceback=False):
         logger.debug(msg)
 
     def start(self):
+        # type: () -> None
         """ Start the web service in a new thread. """
-        thread = threading.Thread(target=self.run)
-        thread.setName("Web service thread")
-        thread.daemon = True
+        thread = threading.Thread(name='WebService runner', target=self.run)
```
khenderick, line 2428: This is not a daemon thread anymore. Expected?

src/openmotics_service.py
```diff
@@ -230,15 +231,17 @@ def stop(signum, frame):
...
             master_controller.stop()
             maintenance_controller.stop()
-            web_service.stop()
             metrics_collector.stop()
             metrics_controller.stop()
             thermostat_controller.stop()
             plugin_controller.stop()
+            communication_led_controller.stop()
             event_sender.stop()
+            watchdog.stop()
```
khenderick, line 244: Stopping the watchdog after certain communicators are stopped might trigger the watchdog and thus unexpected behavior.

src/gateway/daemon_thread.py
LnL7: Just for debugging, startup is rather predictable so didn't need it there.

src/gateway/webservice.py
LnL7: It actually doesn't need a thread at all from what I can tell, any reason to keep it?

src/gateway/metrics_controller.py
LnL7: Already added `DaemonThread.sleep` which waits one the daemon's stop event similar to how the internal loop works.

src/gateway/daemon_thread.py
```diff
@@ -0,0 +1,85 @@
...
+        # type: () -> None
+        backoff = 0
+        while not self._stop.is_set():
+            start = time.time()
+            if not self._parent.is_alive():
+                logger.info('Aborting daemon {}'.format(self._name))
+                return
+            try:
+                self._target()
+                self.sleep(max(0, self._interval - (time.time() - start)))
```
khenderick, line 76: I'd rather not have this in a tight loop but wait e.g. 0.1 seconds

LnL7: There are a few cases where the interval is `0.1` or less, most notably the EventSender. Is it ok to increase those?

PR #298, opened 2020-04-17
Busybox compatibility

khenderick:


src/post-update.sh
```diff
@@ -2,6 +2,8 @@
 set -e
 
 OS_DIST=`awk -F= '$1=="ID" { print $2 ;}' /etc/os-release`
-TEMP_DIR=$(mktemp -d -t ci-XXXXXXXX --tmpdir=/opt/openmotics)
+DATE=`date +%Y%m%d%H%M%S`
+TEMP_DIR="/opt/openmotics/tmp_$DATE"
+mkdir $TEMP_DIR
```
LnL7, line 7: Does `$(TMDPIR=/opt/openmotics/tmp mktemp)` work? Using tmpdir is a bit safer since it ensures uniqueness, but that shouldn't really matter in case.

khenderick:
```
root@OpenMotics:~# $(TMDPIR=/opt/openmotics/tmp mktemp)
-sh: /tmp/tmp.3zjmKc: Permission denied
```

Unfortunately it wants to create a directory in `/tmp`. But I think the date should be "random" enough and I don't expect that code to be run twice at the same second.

khenderick:
Tests fail since jenkins expects `run3.sh` to be available which isn't the case on master yet, so I'll ignore this for now.

PR #300, opened 2020-04-20
Fixed classic master return value for set_virtual_sensor

svanscho:


khenderick:
Let's create a fix on master

PR #303, opened 2020-04-20
Fixed classic master return value for set_virtual_sensor

svanscho:


khenderick:
Python 2 tests: PASS
Python 3 tests: FAIL because the scripts are not yet on `master`
This PR can be merged in

svanscho:
Fails because of run3.sh.

PR #307, opened 2020-04-23
Fix regression in save_sensor(s)

khenderick:


khenderick:
Replaces #305 by @LnL7

khenderick:
Fails as expected due to `run3.sh` not being in `master`, which is fine.

PR #313, opened 2020-04-24
core+ master tool

LnL7:


src/core_updater.py
```diff
@@ -55,7 +55,6 @@ def setup_logger():
     core_cli_serial_port = config.get('OpenMotics', 'cli_serial')
     Injectable.value(cli_serial=Serial(core_cli_serial_port, 115200))
     Injectable.value(core_communicator=None)
-    Injectable.value(maintenance_communicator=None)
```
khenderick, line 58: Hmm, that one is injected, so that should be registered, right?

src/master_tool.py
```diff
@@ -18,21 +18,142 @@
...
+        sys.exit(1)
+
+
+@Inject
+def classic_master_sync(master_communicator=INJECTED):
+    print('Sync...')
+    try:
+        master_communicator.do_command(master_api.status())
+        print('Done sync')
+        sys.exit(0)
```
khenderick, line 67: This seems inconsistent with other methods that don't exit on success.

PR #314, opened 2020-04-27
Rooms in the ORM

khenderick:


src/gateway/migrations/orm/001_init_tables.py
```diff
@@ -20,23 +20,10 @@
...
-except ImportError:
-    pass
-
-SQL = pw.SQL
-
-logger = logging.getLogger('openmotics')
+from gateway.models import (
+    ThermostatGroup, Feature, Output, ValveToThermostat, OutputToThermostatGroup,
+    DaySchedule, Preset, Thermostat, Valve, PumpToValve, Pump
+)
```
svanscho, line 26: FYI: these are changes to an already existing migration in the field.  I don't think there will be any impact schema/db wise but we should validate this.

src/gateway/models.py
```diff
@@ -68,7 +95,37 @@ class Feature(BaseModel):
 class Output(BaseModel):
     id = PrimaryKeyField()
     number = IntegerField(unique=True)
-    # TODO: model AnalagOutput and DigitalOutput and use as dimmer etc
+    room = ForeignKeyField(Room, null=True, on_delete='SET NULL', backref='outputs')
```
svanscho, line 98: pretty sure we want an output type here soon, perhaps add it already? Or will we do this together with names etc?

src/gateway/orm_syncer.py
```diff
@@ -0,0 +1,52 @@
...
+
+from ioc import Inject, INJECTED
+from gateway.hal.master_event import MasterEvent
+from gateway.hal.master_controller import MasterController
+from gateway.models import Output, Room
+
+logger = logging.getLogger("openmotics")
+
+
+class ORMSyncer(object):
```
svanscho, line 30: Perhaps move this under some master class/module?

src/gateway/orm_syncer.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        logger.info('Sync ORM with Master/Core reality')
+
+        # Outputs
+        output_ids = []
+        for output_dto in master_controller.load_outputs():
+            output_id = output_dto.id
+            output_ids.append(output_id)
+            Output.get_or_create(number=output_id)
+        Output.delete().where(Output.number.not_in(output_ids)).execute()
```
svanscho, line 52: This is dangerous in the future as outputs can also be created by plugins. Can we add a source and 'external_id' on the outputs? This will then change into something like
`where external_id == master.output_id and source=='master'`

src/gateway/models.py
svanscho: would rename the number to external_id, as this will be the id as known on e.g. the master and the plugins. Not sure Integer will be sufficient. We might want to use string-based ids, depending on what plugins will give us. Also a source like 'com.openmotics.master' en then 'com.openmotics.plugin.xyz' would be good to already implement (see later in my review).

src/gateway/output_controller.py
```diff
@@ -0,0 +1,66 @@
...
+@Injectable.named('output_controller')
+@Singleton
+class OutputController(object):
+
+    @Inject
+    def __init__(self, master_controller=INJECTED):
+        self._master_controller = master_controller  # type: MasterController
+
+    def load_output(self, output_id):  # type: (int) -> OutputDTO
+        output = Output.get(number=output_id)  # type: Output
```
svanscho, line 40: I believe in the output controller the output_id will correspond to the id field on the database. Number would be the external_id as the id known by the master (see above). Right now the number should be unique. We need a way for plugins to register outputs with potentially the same external_id (right now: number) in the future. The OutputController will work id wise (db id) and we can use the external id to refine the where clauses if we already know which source it is. 

Unless you want to implement the load_output as always coming from the master? Then, in the future, we also need something like: `load_output(source, external_id)`

src/openmotics_service.py
```diff
@@ -221,11 +253,17 @@ def start(master_controller=INJECTED, maintenance_controller=INJECTED,
         thermostat_controller.start()
         metrics_collector.start()
         web_service.start()
-        plugin_controller.start()
         communication_led_controller.start()
         event_sender.start()
         watchdog.start()
 
+        # Always
+        ORMSyncer.sync()
```
svanscho, line 261: Not sure I like this here. Perhaps move this to a background task running as part of the master controller? For the rooms we can probably just do it once as we don't intend on using the rooms on the master in the future, correct? Or do we still need it for all_lights_off per room functionality?

src/gateway/models.py
khenderick: Yes, this model (and similar models) will be changed once we start on letting other components (such as plugins) register entries and when cross-configuration is needed. Let's postpone the changes till those ideas are more clear.

src/gateway/orm_syncer.py
khenderick: Similar to the other comment, this will be adapted once those extra identifiers are present.

src/gateway/output_controller.py
khenderick: Currently the controllers & DTOs have a very tight interface to the classic structure. It will indeed be adapted, but let's do this when the actual changes are made.

src/openmotics_service.py
khenderick: Rooms are indeed not synced. I'll take a look on moving or even removing it. It's already triggered on certain events (eeprom changes, maintenance mode), so I'll try to trigger it once the MasterController is started or similar.

src/gateway/migrations/orm/001_init_tables.py
khenderick: Validated, no impact on changing previously applied migrations as long as the net outcome is identical (meaning the changes applied in `migrate` and/or `rollback`).

src/openmotics_service.py
khenderick: The issue with trying to remove this from the startup sequence is that various lists are now ORM driven. This means that if the hardware setup was changed when the service was down, the ORM won't hold e.g. a set of Outputs. This means that until this sync runs, none of the calls or logic will show these new Outputs. Moreover, since the service is just started, all (e.g. eeprom) caches are clear and won't trigger any update.

So I think during startup is the correct moment to run this sync. And obviously when we actually detect possible changes such as eeprom changes or the exit of maintenance mode.

svanscho: Understand. Can't we move it to the startup of the master controller?

src/openmotics_service.py
```diff
@@ -22,7 +22,9 @@
 import logging
 import time
 import constants
-from models import Database, Feature
+from gateway.models import Database, Feature
+from master.orm_syncer import ORMSyncer
```
LnL7, line 26: This broke the integration tests, that module isn't committed/used.

PR #318, opened 2020-05-05
Fixing import error and "python2 vs python"

khenderick:


src/gateway/metrics_collector.py
```diff
@@ -282,64 +280,72 @@ def _run_system(self, metric_type):
...
-                    for reading in ['available', 'used', 'percent', 'free', 'inactive', 'shared', 'active', 'total']:
-                        try:
-                            key = 'memory_{0}'.format(reading)
-                            value = memory[reading]
-                            values[key] = int(value) if reading != 'percent' else float(value)
-                        except Exception as ex:
-                            logger.error('error loading memory metric: {0}'.format(ex))
-                except Exception as ex:
-                    logger.error('Error loading memory metrics: {0}'.format(ex))
+                if collect_disk_metrics:
```
svanscho, line 289: Isn't this about system metrics rather than only disk metrics? Also I believe disk metrics themselves are the only system metrics that don't rely on psutil.

PR #327, opened 2020-05-14
include events for debug dumps in the healthcheck

LnL7:
This enables tracking of service and master restarts in the cloud
without sending over the enitre debug dump by default.

### DIFF
--- a/src/vpn_service.py
+++ b/src/vpn_service.py
@@ -40,6 +40,9 @@ from ioc import Injectable, INJECTED, Inject
 from bus.om_bus_client import MessageClient
 from bus.om_bus_events import OMBusEvents
 
+if False:  # MYPY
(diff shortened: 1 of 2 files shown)


src/vpn_service.py
```diff
@@ -520,7 +527,14 @@ def _check_vpn(self):
                 dumps = {}
                 dumps.update(self._get_debug_dumps('energy'))
                 dumps.update(self._get_debug_dumps('master'))
-                call_data['debug'] = {'dumps': dumps}
+
+                dump_info = {k: v.get('data', {}).get('action', 'unknown_restart')
```
khenderick, line 531: I think you'll also have to include the "type", otherwise we won't know whether there were issues in master or energy communication.

PR #331, opened 2020-05-18
remove classic/core+ singletons

LnL7:
Having overlapping singleton registrations meant that the classic and
core+ modules couldn't be imported at the same time. This means code
needs to be carefully imported depending on what platform is being
tested, particularly problematic for testing.


svanscho:
So now we can support the Core+ and the classic at the same time! Or multiple Cores :-)

khenderick:
Does this code run/start on both Classic Gateway and Core+?

LnL7:
Tested both now.

PR #339, opened 2020-05-19
unittest coverage

LnL7:


testing/unittests/run.sh
```diff
@@ -6,4 +6,7 @@ cd "$(dirname "${BASH_SOURCE[0]}")/../.."
 # TODO install egglink instead
 export PYTHONPATH=$PYTHONPATH:$PWD/src
 
-pytest testing/unittests --log-level=DEBUG --durations=2 --junit-xml testing/gw-unit-reports/gateway.xml
+pytest testing/unittests --log-level=DEBUG --durations=2 \
+    --junit-xml testing/gw-unit-reports/gateway.xml \
+    --cov-report xml --cov-fail-under=50 \
```
svanscho, line 11: perhaps we could make this configurable in Jenkins?

LnL7: that makes it harder to reproduce locally and would cause problems when we want to increase this (eg. coverage will still be lower on master)

svanscho: ok, increasing it this way will require a code commit but I guess that's fine given the advantages of reproducibility. Every commit will have reproducible unit tests.

PR #340, opened 2020-05-26
Feature/python3 tests

mdemeyer:


src/gateway/maintenance_communicator.py
```diff
@@ -17,7 +17,6 @@
 """
 from __future__ import absolute_import
 
-
```
mdemeyer, line 20: This looks like a bad merge.

src/master/classic/master_communicator.py
```diff
@@ -135,6 +135,7 @@ def __get_cid(self):
         return ret
 
     def __write_to_serial(self, data):
+        # type: (bytearray) -> None
         """ Write data to the serial port.
 
         :param data: the data to write
```
mdemeyer, line 141: Update documentation with changed types!

requirements.txt
```diff
@@ -6,7 +6,7 @@ hypothesis == 4.57.1
 mock == 2.0.0
 pytest == 4.6.9
 pytest-cov == 2.8.1
-cherrypy == 3.2.2
+cherrypy == 17.4.2
```
khenderick, line 9: I'd revert this and `ujson` in this PR for now to reduce the impact scope.

src/master/classic/master_command.py
```diff
@@ -46,12 +48,13 @@ def __init__(self, action, input_fields, output_fields, output_action=None):
         :param output_action: name of the action of the answer, as described in the Master api. None if identical to the mainaction
         :type output_action: 2-byte string
         """
-        self.action = action
+        self.action = bytearray(action.encode())
         self.input_fields = input_fields
         self.output_fields = output_fields
-        self.output_action = action if output_action is None else output_action
+        self.output_action = bytearray(action.encode()) if output_action is None else bytearray(output_action.encode())
```
khenderick, line 54: Use `... = self.action if output_action is ...` instead

src/master/classic/master_command.py
```diff
@@ -46,12 +48,13 @@ def __init__(self, action, input_fields, output_fields, output_action=None):
         :param output_action: name of the action of the answer, as described in the Master api. None if identical to the mainaction
         :type output_action: 2-byte string
```
khenderick, line 49: Can you replace the `:type ...` lines with Mypy typings?

src/master/classic/master_command.py
```diff
@@ -77,16 +79,17 @@ def create_input(self, cid, fields=None, extended_crc=False):
...
     def __calc_crc(encoded_string):
+        # type: (bytearray) -> bytearray
         """ Calculate the crc of an string. """
         crc = 0
         for byte in encoded_string:
-            crc += ord(byte)
+            crc += byte
 
-        return 'C' + chr(crc // 256) + chr(crc % 256)
+        return bytearray(b'C') + bytearray([crc // 256]) + bytearray([crc % 256])
```
khenderick, line 92: Use `return bytearray(b'C') + bytearray([crc // 256, crc % 256])`

src/master/classic/master_command.py
```diff
@@ -96,11 +99,15 @@ def create_output(self, cid, fields):
         :type cid: byte
         :param fields: dictionary with values for the fields
         :type fields: dict
-        :rtype: string
+        :rtype: bytearray
         """
-        ret = self.output_action + chr(cid)
+        ret = self.output_action + bytearray(chr(cid).encode())
```
khenderick, line 104: I think here you can just use `... + bytearray([cid])`

src/master/classic/master_command.py
```diff
@@ -96,11 +99,15 @@ def create_output(self, cid, fields):
...
         for field in self.output_fields:
-            ret += field.encode(fields.get(field.name))
+            encoded = field.encode(fields.get(field.name))
```
khenderick, line 106: I think it would make more sense if we make sure that `.encode(...)` always returns a `bytearray`. this wil increase consistency :)

src/master/classic/master_command.py
```diff
@@ -327,23 +334,24 @@ def encode(self, field_value):
             if field_value < 0 or field_value > 255:
                 raise ValueError('Int does not fit in byte: %d' % field_value)
             else:
-                return chr(field_value)
+                return bytearray([field_value % 256])
```
khenderick, line 337: I don't think the `% 256` is needed as the value is checked anyway

src/master/classic/master_command.py
```diff
@@ -415,7 +423,7 @@ def get_min_decode_bytes(self):
 
     def decode(self, byte_str):
         """ Checks if byte_str is the literal """
-        if byte_str != self.literal:
+        if byte_str != bytearray(self.literal.encode()):
```
khenderick, line 426: I think it would be better to use `self.literal = literal.encode()` in the constructor so it always returns a `bytearray` in `encode()`

src/serial_utils.py
```diff
@@ -37,9 +37,12 @@ def printable(data):
     if isinstance(data, list) or isinstance(data, bytearray):
         byte_notation = ' '.join(['{0: >3}'.format(i) for i in data])
         string_notation = ''.join([str(chr(i)) if 32 < i <= 126 else '.' for i in data])
-    else:
+    elif isinstance(data, str):
         byte_notation = ' '.join(['{0: >3}'.format(ord(c)) for c in data])
         string_notation = ''.join([c if 32 < ord(c) <= 126 else '.' for c in data])
+    else:
```
khenderick, line 43: I think this change can be reverted, `printable(...)` already supports `bytearray`.

testing/unittests/master_tests/master_command_test.py
```diff
@@ -50,9 +50,9 @@ def test_encode_byte_field(self):
 
     def test_decode_byte_field(self):
         """ Test for Field.byte.decode """
-        self.assertEqual(0, Field.byte("test").decode('\x00'))
-        self.assertEqual(1, Field.byte("test").decode('\x01'))
-        self.assertEqual(255, Field.byte("test").decode('\xFF'))
+        self.assertEqual(0, Field.byte("test").decode(bytearray(b'\x00')))
```
khenderick, line 53: Can you compare with `bytearray([0])` etc here? As this makes more sense since it's comparing a byte to a byte instead of a byte to an encoded value. It's easier to read `255` than `\xFF`

khenderick: Similar remarks in the code below where integers are compared

src/master/classic/master_command.py
mdemeyer: I don't think it is possible to specify the "2-byte" length part of the type.
As an alternative I used assert(len == 2).

khenderick:
Most of these changes are already merged with other PRs in smaller chunks based on certain logic.

PR #345, opened 2020-05-29
P1 metrics

LnL7:


src/gateway/gateway_api.py
```diff
@@ -777,48 +783,41 @@ def get_realtime_power(self):
...
 
                 out = []
                 for i in range(num_ports):
-                    out.append([convert_nan(volt[i], default=0.0), convert_nan(freq[i], default=0.0),
-                                convert_nan(current[i], default=0.0), convert_nan(power[i], default=0.0)])
+                    data = {'voltage': convert_nan(volt[i], default=0.0),
+                            'frequency': convert_nan(freq[i], default=0.0),
+                            'current': convert_nan(current[i], default=0.0),
+                            'power': convert_nan(power[i], default=0.0)}
+                    out.append(RealtimePower(**data))
```
khenderick, line 820: Wouldn't it be more clear to use

```
out.append(RealtimePower(voltage=...,
                         frequency=...,
                         ...))
```

src/gateway/gateway_api.py
```diff
@@ -828,38 +827,50 @@ def get_realtime_power(self):
...
+
     def get_total_energy(self):
+        # type: () -> Dict[str,List[List[Optional[float]]]]
         """ Get the total energy (kWh) consumed by the power modules.
 
         :returns: dict with the module id as key and the following array as value: [day, night].
         """
         output = {}
-        if self.__power_communicator is None or self.__power_controller is None:
+        if self.__power_controller is None:
```
khenderick, line 849: Also check `self.__power_store`

src/gateway/gateway_api.py
```diff
@@ -828,38 +827,50 @@ def get_realtime_power(self):
...
+        modules = self.__power_store.get_power_modules()
         for module_id in sorted(modules.keys()):
             try:
-                addr = modules[module_id]['address']
                 version = modules[module_id]['version']
                 num_ports = power_api.NUM_PORTS[version]
 
-                day = [None] * num_ports  # type: List[Optional[int]]
-                night = [None] * num_ports  # type: List[Optional[int]]
+                day = [None] * num_ports  # type: List[Optional[float]]
```
khenderick, line 858: These are counters and are not supposed to be floats.

src/gateway/metrics_collector.py
```diff
@@ -611,82 +616,120 @@ def _run_pulsecounters(self, metric_type):
...
+        except CommunicationTimedOutException:
+            logger.error('Error getting realtime power: CommunicationTimedOutException')
+        except InMaintenanceModeException:
+            logger.info('Error getting realtime power: InMaintenanceModeException')
+        except Exception as ex:
+            logger.exception('Error getting realtime power: {0}'.format(ex))
+        try:
+            realtime_p1_data = self._gateway_api.get_realtime_p1()
+            for module_id, device_id in mapping.items():
+                for realtime_p1 in realtime_p1_data:
```
khenderick, line 668: Wouldn't it be cleaner to make `realtime_p1_data` a dict per module_id?

src/gateway/metrics_collector.py
```diff
@@ -611,82 +616,120 @@ def _run_pulsecounters(self, metric_type):
...
+                                                          'electricity_voltage_phase2': electricity_p1['voltage']['phase2'],
+                                                          'electricity_voltage_phase3': electricity_p1['voltage']['phase3'],
+                                                          'electricity_current_phase1': electricity_p1['current']['phase1'],
+                                                          'electricity_current_phase2': electricity_p1['current']['phase2'],
+                                                          'electricity_current_phase3': electricity_p1['current']['phase3']},
+                                                  tags={'type': 'openmotics',
+                                                        'id': device_id.format(realtime_p1['port_id']),
+                                                        'ean': electricity_p1['ean']},
+                                                  timestamp=now)
+                        gass_p1 = realtime_p1.get('gass', {})
```
khenderick, line 686: That should be `gas` right?

src/power/power_controller.py
```diff
@@ -13,159 +13,296 @@
...
+            payloads['phase{}'.format(i)] = self._power_communicator.do_command(module['address'], cmd)[0]
+
+        currents = []
+        for port_id in range(NUM_PORTS[P1_CONCENTRATOR]):
+            phases = {}
+            for phase, payload in payloads.items():
+                try:
+                    phases[phase] = float(payload[port_id * 5:(port_id + 1) * 6][:3])
+                except ValueError:
+                    phases[phase] = 0.0
```
khenderick, line 233: Not sure if this is a correct approach. I would rather assume this key would not exist or the value would be None in case of a single-phase meter. Then we can differentiate between a 3 phase meter that has phases disconnected (value = 0.0) or a single phase meter (value = None).

src/gateway/gateway_api.py
LnL7: Not sure what you mean. Those values where floats already, just not strictly typechecked yet.

LnL7: n/m

src/power/power_controller.py
LnL7: Updated, but only for the new P1 metrics. The existing get_realtime_power/get_total_energy still fallback to 0, I'm not sure what the correct thing to do there is.

PR #353, opened 2020-06-10
generalize module types in integration tests

LnL7:


src/gateway/hal/master_controller_classic.py
```diff
@@ -928,7 +935,8 @@ def get_configuration_dirty_flag(self):
     def _update_modules(self, api_data):
         # type: (Dict[str,Any]) -> None
         """ Create a log entry when the MI message is received. """
-        module_map = {'O': 'output', 'I': 'input', 'T': 'temperature', 'D': 'dimmer'}
+        logger.debug('update modules {}'.format(api_data))
```
khenderick, line 938: I'd remove this raw logging here. However, I think it's a good idea to add some "clean" logging (e.g. "New output module found") with info level below.

testing/cicd/tests/discover_test.py
```diff
@@ -0,0 +1,50 @@
...
+    finally:
+        toolbox.module_discover_stop()
+
+
+@pytest.mark.smoke
+def test_output_module(toolbox, discover_mode):
+    toolbox.tester.toggle_output(toolbox.DEBIAN_DISCOVER_OUTPUT)
+    toolbox.assert_modules(1)
+    data = toolbox.dut.get('/get_module_log')['log']
+    assert ['WARN', 'Existing output module found.'] in data
```
khenderick, line 34: Is this a consistent test? I would expect that depending on when this test runs it could either be an existing or a new module?

LnL7: Yes, since initialisation always ensures modules are discovered first. Ideally testing factory reset / discovery would be more separate from the rest of the tests.

PR #354, opened 2020-06-10
Added output locks

svanscho:


src/gateway/hal/master_controller_classic.py
```diff
@@ -1233,3 +1270,44 @@ def save_pulse_counters(self, pulse_counters):  # type: (List[Tuple[PulseCounter
     def get_pulse_counter_values(self):  # type: () -> Dict[int, int]
         out_dict = self._master_communicator.do_command(master_api.pulse_list())
         return {i: out_dict['pv{0}'.format(i)] for i in range(24)}
+
+    # Validation bits
+
+    def load_validationbits(self):
```
svanscho, line 1276: @khenderick this one will need an update based on Wilbert's latest master hex. The bitshift code will be needed and is provided in the comments, so you can just use that one.

svanscho: I very explicitly kept the validation_bit as output configuration instead of a 'locked' bit. Once we have a "validationbit-compiler" we can then use it in a more abstract manner. For now the local GW API will allow to set a validation bit. The cloud however does not implement this (yet). One thing to check is if the synchronization still works.

src/master/classic/outputs.py
```diff
@@ -34,35 +40,35 @@ def __init__(self, on_output_change=None):
...
-            for output_id, output in six.iteritems(self._outputs):
-                self._update_maybe_report_change(output, {'status': output_id in on_dict,
-                                                          'dimmer': on_dict.get(output_id)})
+        for output_id, output in six.iteritems(self._outputs):
+            if output_id in on_dict:
+                on_output = on_dict[output_id]
+                self._update(output_id, on_output)
+            else:
+                off_output = output
+                output['status'] = False
```
khenderick, line 57: This updates the same object referenced in the `self._outputs` list and thus the `self._update(...)` won't detect any changes.

khenderick:
Taking over this ticket due to @svanscho's holiday.

PR #355, opened 2020-06-11
Add mypy types to all files in power module

mdemeyer:


src/gateway/gateway_api.py
```diff
@@ -957,13 +957,13 @@ def get_energy_time(self, module_id, input_id=None):
             input_ids = [input_id]
         data = {}
         for input_id in input_ids:
-            voltage = list(self.__power_communicator.do_command(addr, power_api.get_voltage_sample_time(version), input_id, 0))
-            current = list(self.__power_communicator.do_command(addr, power_api.get_current_sample_time(version), input_id, 0))
-            for entry in self.__power_communicator.do_command(addr, power_api.get_voltage_sample_time(version), input_id, 1):
+            voltage = list(self.__power_communicator.do_command(addr, power_api.get_voltage_sample_time(version), input_id, "0"))
```
LnL7, line 960: why are the list conversions needed and is the behaviour with "0" equivalent?

mdemeyer: On line 965 voltage is used as a list, so we need the conversion. It is not possible to append to a Tuple.
I don't think the behavior with "0" is the same. I reverted that and expanded the types accepted by do_command.

PR #366, opened 2020-06-23
OutputController refactoring

LnL7:


src/gateway/gateway_api.py
```diff
@@ -209,22 +212,22 @@ def flash_leds(self, led_type, led_id):
 
     # Output functions
 
-    def get_outputs_status(self):  # type: () -> List[Dict[str, Any]]
+    def get_output_statuses(self):
```
khenderick, line 215: These should be removed, callers should directly call the OutputController

src/gateway/metrics_collector.py
```diff
@@ -447,17 +447,17 @@ def _run_system(self, metric_type):
...
     def _run_outputs(self, metric_type):
+        # type: (str) -> None
         while not self._stopped:
             start = time.time()
             try:
-                result = self._gateway_api.get_outputs_status()
-                for output in result:
-                    output_id = output['id']
-                    if output_id not in self._environment_outputs:
+                result = self._gateway_api.get_output_statuses()
```
khenderick, line 454: Directly use the OutputController instead

PR #367, opened 2020-06-24
Support tight loops in DaemonThread

khenderick:
This will make sure that metrics are distributed fast enough, preventing memory leakage.

### DIFF
--- a/src/gateway/daemon_thread.py
+++ b/src/gateway/daemon_thread.py
@@ -21,7 +21,7 @@ import time
 logger = logging.getLogger("openmotics")
 
 if False:  # MYPY
-    from typing import Any, Callable
+    from typing import Any, Callable, Optional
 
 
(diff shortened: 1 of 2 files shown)


src/gateway/daemon_thread.py
```diff
@@ -77,10 +79,12 @@ def _run(self):
                 return
             try:
                 self._target()
-                self.sleep(max(0.1, self._interval - (time.time() - start)))
+                if self._interval > 0:
+                    min_wait_time = 0.1 if self._interval >= 0.5 else 0.05
```
LnL7, line 83: This was mostly there to prevent busy loops, so just changing the default to 0.05 would be fine.

khenderick: I think this is reasonable, will lower the CPU usage a little bit, and 0.1 is perfect for slower loops.

PR #371, opened 2020-07-02
Refactor update logic to prevent various issues

khenderick:
Fixed issues:
* Lockup during update
* Ugly stacktraces during interpreter shutdown
* Use of ugly shell script to flash the classic master
* Inconsistent logging
* Minimal running code during update


src/master_tool.py
```diff
@@ -23,39 +23,46 @@
...
-    logger.setLevel(logging.INFO)
+    logger.setLevel(logging.DEBUG)
     logger.propagate = False
 
     handler = logging.StreamHandler()
     handler.setLevel(logging.INFO)
     handler.setFormatter(logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s"))
     logger.addHandler(handler)
 
+    handler = handlers.RotatingFileHandler(constants.get_update_log_location(), maxBytes=3 * 1024 ** 2, backupCount=2)
```
LnL7, line 57: Won't this include logging for everything?

khenderick: It will include logging for all `master_tool.py` CLI runs. So strictly speaking a casual `master_tool.py --reset` or similar will also log to that file, but that's fine.

PR #381, opened 2020-07-07
master/slave firmware update

LnL7:


src/gateway/gateway_api.py
```diff
@@ -141,6 +141,10 @@ def get_main_version(self):
     def reset_master(self):
         return self.__master_controller.cold_reset()
 
+    def update_firmware(self, firmware_file):
```
svanscho, line 144: can we rename this to update_master? firmware is also used on slave modules... 
other option would be to call this update_module and then provide a type like 'master', 'output', or some other type identifier.

Also, do we need this on the gateway_api class? Idea originally was to split up the gateway_api into seperate classes (like e.g. for the output controller).

src/gateway/webservice.py
```diff
@@ -2043,6 +2044,24 @@ def get_update_output(self):
         return {'output': output,
                 'version': version}
 
+    @openmotics_api(auth=True, plugin_exposed=False)
+    def update_firmware(self, md5, firmware_data):
```
svanscho, line 2048: same comment as above. I do like the extra module_type kwarg approach but let's discuss

svanscho: as I was thinking about it, we could then also use it for updating the gw code itself.

src/gateway/webservice.py
```diff
@@ -488,6 +494,45 @@ def reset_master(self):
...
+        hasher.update(firmware_data)
+        calculated_md5 = hasher.hexdigest()
+        if md5 != calculated_md5:
+            raise ValueError('firmware md5:%s does not match' % calculated_md5)
+
+        temp_file = '/tmp/{}.hex'.format(md5)
+        with open(temp_file, 'wb') as firmware_file:
+            firmware_file.write(firmware_data)
+        self._gateway_api.update_master_firmware(temp_file)
+        return {}
```
khenderick, line 513: The file should be cleaned up. In a best-case scenario the Gateway never reboots, and updates could eventually fill up the /tmp filesystem. Or maybe move it to the location where the classic update also leaves the currently installed firmware.

src/master/classic/master_communicator.py
```diff
@@ -91,34 +99,50 @@ def __init__(self, controller_serial=INJECTED, init_master=True, verbose=False,
...
+    def update_mode_start(self):
+        # type: () -> None
+        if self.__maintenance_mode:
+            raise InMaintenanceModeException()
+        self.__update_mode = True
+
+    def update_mode_stop(self):
+        # type: () -> None
+        self.__update_mode = False
+        self._flush_serial_input()
```
khenderick, line 125: Wouldn't it be corrected to swap these two lines? So that the `_flush_serial_input` can to it's work before the `__read` thread is released again?

PR #387, opened 2020-07-13
cloud firmware updates

LnL7:


src/gateway/webservice.py
```diff
@@ -2091,6 +2052,86 @@ def get_update_output(self):
...
+        config = ConfigParser()
+        config.read(constants.get_config_file())
+        cloud_url = config.get('OpenMotics', 'cloud_url')
+        gateway_uuid = config.get('OpenMotics', 'uuid')
+
+        url_template = cloud_url + '/portal/firmware_metadata/{}/{}/?uuid=' + gateway_uuid
+        if master:
+            temp_file = self._download_firmware(url_template.format('master_classic', master))
+            self._gateway_api.update_master_firmware(temp_file)
+            shutil.move(temp_file, '/opt/openmotics/firmware.hex')
```
svanscho, line 2066: do we want to store some metadata in the filename or will we be able to resolve this using the sha256?

LnL7: This is mostly for rollback AFAIK. For debugging purposes the hash and/or version the master reports should be enough I think.

src/gateway/webservice.py
```diff
@@ -2091,6 +2052,86 @@ def get_update_output(self):
         return {'output': output,
                 'version': version}
 
+    @openmotics_api(auth=True, plugin_exposed=False)
+    def update_firmware(self, master=None, can=None):
+        config = ConfigParser()
+        config.read(constants.get_config_file())
+        cloud_url = config.get('OpenMotics', 'cloud_url')
+        gateway_uuid = config.get('OpenMotics', 'uuid')
```
svanscho, line 2060: would assume this is available somewhere in memory too...

PR #401, opened 2020-08-03
Master classic python 3

khenderick:


src/gateway/hal/master_controller_classic.py
```diff
@@ -1144,7 +1137,7 @@ def get_backup(self):
                 retry = bank
                 logger.warning('Got timeout reading bank {0}. Retrying...'.format(bank))
                 time.sleep(2)  # Doing heavy reads on eeprom can exhaust the master. Give it a bit room to breathe.
-        return output
+        return ''.join(chr(c) for c in output)
```
LnL7, line 1140: perhaps `output.decode('ascii')` is a bit nicer here

src/master/classic/eeprom_controller.py
```diff
@@ -750,13 +727,15 @@ def get_name(self):
...
+        # type: (bytearray) -> str
+        dirty = remove_tail(data)
+        return ''.join(chr(i) if i < 128 else ' ' for i in dirty)
 
     def encode(self, field):
+        # type: (Optional[str]) -> bytearray
         if field is None:
             field = ''
-        return append_tail(field, self._length)
+        return append_tail(bytearray([ord(c) for c in field]), self._length)
```
LnL7, line 738: perhaps bytearray(field.encode('ascii')) is a bit nicer here, same in a few other places

src/master/classic/eeprom_controller.py
```diff
@@ -902,13 +891,15 @@ def get_name(self):
...
 
     def encode(self, field):
-        actions = ''
+        # type: (Optional[str]) -> bytearray
+        actions = bytearray()
         if field is not None and len(field) > 0:
-            actions = ''.join([chr(int(x)) for x in field.split(",")])
-        return append_tail(actions, self._length, '\xff')
+            actions = bytearray([int(x) for x in field.split(',')])
+        return append_tail(actions, self._length)
```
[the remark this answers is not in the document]
khenderick: Yes, it's the default argument

src/gateway/hal/master_controller_classic.py
khenderick: Unfortunately that doesn't work for certain characters:

```
b = bytearray([i for i in range(256)])
b.decode('ascii')
> UnicodeDecodeError: 'ascii' codec can't decode byte 0x80 in position 128: ordinal not in range(128)
```

src/master/classic/eeprom_controller.py
khenderick: That seems not possible;

```
s = 'abc\x00\xff'
bytearray(s.encode('ascii'))
> UnicodeDecodeError: 'ascii' codec can't decode byte 0xff in position 4: ordinal not in range(128)
```

PR #403, opened 2020-08-05
Version 2 of output status decorator

svanscho:


src/plugin_runtime/runtime.py
```diff
@@ -75,42 +77,45 @@ def _init_plugin(self):
...
+        for decorator_name, target_receivers in six.iteritems(receiver_mapping):
+            for decorated_method, decorator_version in get_special_methods(self._plugin, decorator_name):
+                target_receivers.append(decorated_method)
+                # this set is used for checking which decorators and its versions are used in the plugin
+                registered_receiver = (decorator_name, decorator_version)  # sets can't take dicts
+                receivers.add(registered_receiver)
 
-            if len(target) > 0:
-                self._receivers.append(method_attribute)
+        self._receivers = [{'decorator_name': decorator_name, 'decorator_version': decorator_version} for (decorator_name, decorator_version) in receivers]
```
khenderick, line 89: Switching from a `set` to a `list` will have issues with Mypy as soon as it would start checking this code.

src/plugin_runtime/runtime.py
```diff
@@ -75,42 +77,45 @@ def _init_plugin(self):
...
 
         # Set the exposed methods
-        for method in get_special_methods(self._plugin, 'om_expose'):
-            self._exposes.append({'name': method.__name__,
-                                  'auth': method.om_expose['auth'],
-                                  'content_type': method.om_expose['content_type']})
+        for decorated_method, decorator_version in get_special_methods(self._plugin, 'om_expose'):
```
khenderick, line 92: Decorator version isn't used here and in the below changes. Let's at least add it to the metadata or it will most likely be removed in a subsequent change as it's a soft-warning about non-used variables. Also, as soon as a version is supported for a decorator, we must block any version that's not yet expected to prevent users from setting a version that might break in the future anyway.

src/plugin_runtime/runtime.py
```diff
@@ -75,42 +77,45 @@ def _init_plugin(self):
...
+                                           'metric_type': decorated_method.om_metric_receive['metric_type'],
+                                           'interval': decorated_method.om_metric_receive['interval']})
 
     def _start_background_tasks(self):
         """ Start all background tasks. """
-        tasks = get_special_methods(self._plugin, 'background_task')
-        for task in tasks:
-            thread = Thread(target=PluginRuntime._run_background_task, args=(task,))
-            thread.name = 'Background thread ({0})'.format(task.__name__)
+        for decorated_method, decorator_version in get_special_methods(self._plugin, 'background_task'):
```
khenderick, line 116: As you can specify a version, you should at least block the use of any other version than 1. Otherwise people might (by accident) use a future version that would at this moment work fine and then break in the future. And that's exactly what we're trying to prevent.

src/plugin_runtime/runtime.py
```diff
@@ -143,7 +149,14 @@ def process_stdin(self):
                 elif action == 'input_status':
                     ret = self._handle_input_status(command['event'])
                 elif action == 'output_status':
-                    ret = self._handle_output_status(command['status'])
+                    if action_version == 1:
```
khenderick, line 152: I would just pass `command` completely to the `_handle_output_status` for cleaness. Otherwise we need the if-statements comparing versions both here and in the `_handle_output_status` implementation.

src/plugin_runtime/runtime.py
```diff
@@ -266,7 +291,7 @@ def _handle_request(self, method, args, kwargs):
             return {'success': False, 'exception': str(exception), 'stacktrace': traceback.format_exc()}
 
     def _handle_remove_callback(self):
-        for method in get_special_methods(self._plugin, 'on_remove'):
+        for method, decorator_version in get_special_methods(self._plugin, 'on_remove'):
```
khenderick, line 294: Unused variable

src/plugin_runtime/runtime.py
svanscho: This was a list before...

khenderick: Ah, nevermind, it is `receivers` that is created as a set, not `self._receivers`. My bad :slightly_smiling_face:

src/plugin_runtime/runtime.py
svanscho: good idea

src/plugin_runtime/runtime.py
svanscho: I removed it from om_expose, we don't need versions there now.

src/plugin_runtime/runtime.py
svanscho: Fixed by refactoring, the plugin will now fail on init.

src/plugin_runtime/runtime.py
svanscho: Not using them anymore

src/plugin_runtime/runtime.py
khenderick: Use `_` for the unused parameter

src/plugin_runtime/decorators.py
```diff
@@ -97,14 +78,23 @@ def shutter_status(method):
     Important! This method should not block, as this will result in an unresponsive system.
     Please use a separate thread to perform complex actions on shutten status messages.
     """
-    args, varargs, kwargs, _ = inspect.getargspec(method)
-    args.pop(0)
+    def add_detail(method):
```
khenderick, line 81: The whole `add_detail` was the first implementation of what is now the version. If you're introducing version to this call, I'd rather use version instead of this `add_detail`.

To be backwards compatible, I'd use the default `version=1` parameter, but change `version` to `2` if and only if the `version` is `1` _and_ this method (which could be called `needs_detail`) detects that different arguments are needed.

If in the future we again add some changed, we'd call that version 3.

This way it works backwards compatible, and plugins can (and should) start just specifying `version=2` explicitly so code can be cleaner as you don't need the `add_detail` key anymore at any other place but only the version.

src/plugin_runtime/runtime.py
```diff
@@ -198,39 +221,71 @@ def delayed_stop():
...
+                IO._log_exception('output status', error)
 
     def _handle_shutter_status(self, status):
-        for receiver in self._shutter_status_receivers:
-            if receiver.shutter_status['add_detail']:
-                IO._with_catch('shutter status', receiver, [status['status'], status['detail']])
+        for receiver in self._decorated_methods['shutter_status']:
+            decorator_version = receiver.shutter_status.get('version', 1)
+            if decorator_version == 1:
+                if receiver.shutter_status['add_detail']:
```
khenderick, line 261: Referring to the other comment, this could then just be cleanly rewritten by version 1 and version 2.

PR #414, opened 2020-08-20
Cleanup scheduling

khenderick:


src/gateway/scheduling.py
```diff
@@ -50,7 +43,8 @@ class CommunicationTimedOutException(Exception):  # type: ignore
 
 class Schedule(object):
 
-    timezone = None
+    NO_NTP_LOWER_LIMIT = 1546300800.0  # 2019-01-01
+    TIMEZONE = None
```
yclybouw, line 47: Are the values `self.start` and `self.end` in UTC or in the Timezone. If the latter, then the logic comparing with `time.time()` will be wrong if Timezone is not None.

src/gateway/scheduling.py
```diff
@@ -73,28 +70,27 @@ def is_due(self):
...
         if self.end is not None:
-            return self.start + self.end < time.time()
+            return self.end < time.time()
         return False
 
+    def _set_next_execution(self):
+        base_time = max(Schedule.NO_NTP_LOWER_LIMIT, self.start, time.time())
+        cron = croniter(self.repeat,
+                        datetime.fromtimestamp(base_time,
+                                               pytz.timezone(Schedule.TIMEZONE)))
```
yclybouw, line 91: According to the docs, when `Schedule.TIMEZONE` is None, `datetime.fromtimestamp` will return a local timestamp. Is this wanted?

src/gateway/scheduling.py
```diff
@@ -231,21 +222,27 @@ def stop(self):
 
     def _process(self):
         while self._stop is False:
-            for schedule in self._schedules.values():
+            for schedule_id in list(self._schedules.keys()):
```
yclybouw, line 225: Why not doing `for schedule_id, schedule in self._schedules.items()`?

testing/unittests/gateway_tests/scheduling_test.py
```diff
@@ -224,31 +226,66 @@ def test_two_actions(self):
...
+    def _wait_for_completed(self, schedule, timeout=5):
+        def _is_completed():
+            return schedule.last_executed is not None and schedule.status == 'COMPLETED'
+
+        _time = SchedulingControllerTest.ORIGINAL_TIME
+        _sleep = SchedulingControllerTest.ORIGINAL_SLEEP
+        end = _time() + timeout
+        while not _is_completed() and _time() < end:
+            _sleep(0.01)
+        self.assertTrue(_is_completed())
```
yclybouw, line 290: If would add a specific time zone test. Example: a daily schedule at 07:00 in Europe/Brussels. On 01/01/2020, the next_schedule is 06:00 UTC and on 01/07/2020, the next_schedule is 05:00 UTC.

src/gateway/scheduling.py
khenderick: `Schedule.TIMEZONE` is filled as soon as the `ScheduleController` starts, so it would be unexpected for it to be `None`.

src/gateway/scheduling.py
khenderick: In python3, `.items()` is a generator that will fail (changing dict during iteration) when a schedule would be added/removed

yclybouw: Ah, indeed, this runs asynchronous. Well done :-)

src/gateway/scheduling.py
yclybouw: Response of Kenneth: self.start/end are in Unix timestamps, based on UTC.

testing/unittests/gateway_tests/scheduling_test.py
yclybouw: Not relevant anymore, as the timestamps are all UTC, and this would only test the `croniter` library.

yclybouw (approved):

PR #417, opened 2020-08-26
ensure openmotics.conf can't corrupt

LnL7:
With updates that require a reboot it's possible that recent file
changes are not fully flushed to disk yet. This can result in an empty
config file when it's updated in place.


svanscho (approved):

khenderick:
@LnL7 if this is a hotfix that must be released, can you also bump the version?

LnL7:
Not necessarily, but we should probably include it in the next release if that doesn't include the rework yet.

svanscho (approved):

PR #423, opened 2020-09-02
Feature/migrate user controller

ptuytsch:


src/gateway/user_controller.py
```diff
@@ -0,0 +1,235 @@
...
+                    logger.debug("User login: Could not find any user with name [{}]".format(user.username))
+                    return False, 'invalid_credentials'
+
+                if user_orm.accepted_terms == UserController.TERMS_VERSION:
+                    return True, self._gen_token(user_orm.username, time.time() + timeout)
+                if accept_terms is True:
+                    user_orm.accepted_terms=UserController.TERMS_VERSION
+                    user_orm.save()
+                    return True, self._gen_token(user_orm.username, time.time() + timeout)
+                logger.debug("User login: User [{}] has not accepted the latest terms".format(user.username))
```
khenderick, line 180: If you want to put delimiters around a logged value, use backticks or quotes instead. The `[` and `]` looks like a list and might be confusing.

src/gateway/user_controller.py
```diff
@@ -0,0 +1,235 @@
...
+        else:
+            timed_out = self._tokens[token][1] >= time.time()
+            return timed_out
+
+    def _validate(self, user):
+        # type: (User) -> None
+        """ Checks if the user object is a valid object to store """
+        if user.username is None or not isinstance(user.username, six.string_types) or user.username.strip() == '':
+            raise RuntimeError("A user must have a username")
+        if user.password is None or not isinstance(user.password, six.string_types) or user.password.strip() == '':
```
khenderick, line 230: From what I see the password in the ORM object passed in here is already hashed,so will never be anything else but a filled string. If we want to protect against no or empty passwords we'll need to do it earlier.

src/gateway/webservice.py
```diff
@@ -421,7 +421,13 @@ def login(self, username, password, accept_terms=None, timeout=None):
...
+        user_dto = UserDTO(
+            username=username,
+            password=password,
+            role="admin",
+            enabled=True
+        )
+        success, data = self._user_controller.login([(user_dto, accept_terms, timeout)])
         if success is True:
             return {'token': data}
         if data == 'terms_not_accepted':
```
khenderick, line 433: Here you can use the enums as well.

src/gateway/webservice.py
```diff
@@ -2352,7 +2365,14 @@ def get_metric_definitions(self, source=None, metric_type=None):
...
     @openmotics_api(check=types(confirm=bool), auth=True, plugin_exposed=False)
     def factory_reset(self, username, password, confirm=False):
-        success, _ = self._user_controller.login(username, password)
+        user_dto_login = UserDTO(
+            username=username,
+            password=password,
+            role="admin",
+            enabled=True
+        )
+        users_to_login = [(user_dto_login, True, 3600.0)]
```
khenderick, line 2374: I would not automatically accept the terms here. Let's assume only a fully compliant user can execute a factory reset.

khenderick (changes requested):
Still some loose ends here and there. I also think I forgot to mention our code style guides, so let's check these as well.

src/gateway/enums.py
```diff
@@ -29,3 +29,11 @@ class State(object):
         STOPPED = 'stopped'
         UP = 'up'
         DOWN = 'down'
+
+class UserEnums(object):
+    class AuthenticationErrors(object):
+        INVALID_CREDENTIALS = 'Invalid_credentials'
```
khenderick, line 35: Lowercase `i`.

src/gateway/user_controller.py
```diff
@@ -45,80 +45,76 @@ class UserController(object):
...
-        :type token_timeout: string
+        # type: (Dict[str, str], int) -> None
+        """ 
+        Constructor a new UserController.
         """
-        logger.info("Initializing the user controller")
         self._config = config
         self._token_timeout = token_timeout
-        self._tokens = {} #type: Dict[str, Tuple(str, float)]
+        self._tokens = {} #type: Dict[str, Tuple[str, float]]
```
khenderick, line 54: Comments should have two spaces before the `#`, and one space after. Mypy will ignore this statement (which is the reason why the `(` vs `[` didn't fail the tests previously)

src/gateway/user_controller.py
```diff
@@ -45,80 +45,76 @@ class UserController(object):
...
+        self.save_users(users=[(cloud_user_dto, ['username', 'password', 'accepted_terms'])])
+
+
+    def save_user(self, user_dto, fields):
+        # type: (UserDTO, List[str]) -> None
+        """
+        Saves one instance of a user with the defined fields in param fields
+        """
+        user_orm = UserMapper.dto_to_orm(user_dto, fields)
+        logger.debug("user_orm: {}".format(vars(user_orm)))
```
khenderick, line 71: Can you remove this logging statement?

src/gateway/user_controller.py
```diff
@@ -45,80 +45,76 @@ class UserController(object):
...
         # set username to lowercase to compare on username
-        username = username.lower()
+        username = user_dto.username.lower()
 
         # check if the removed user is not the last admin user of the system
-        if self.get_role(username) == "admin" and self._get_num_admins() == 1:
-            raise Exception("Cannot delete last admin account")
+        if self.get_number_of_users() <= 1:
+            raise Exception(UserEnums.DeleteErrors.LAST_ACCOUNT)
         else:
```
khenderick, line 118: You can remove this else-statement and unindent the below code. Keeps the code a bit more compact

src/gateway/user_controller.py
```diff
@@ -130,92 +126,72 @@ def remove_user(self, username):
...
+                timeout = None
+        if timeout is None:
+            timeout = self._token_timeout
+
+        for user_orm in User.select().where(
+            User.username == user_dto.username.lower(),
+            User.password == user_dto.hashed_password
+        ):
+            # check if returned object is valid
+            if user_orm is None or not hasattr(user_orm, 'username'):
```
khenderick, line 150: Will this ever happen? Since you're looping over all found users, this will only yield existing records.

Aside from this remark. Wouldn't it be more elegant to use something like `user_orm = User.select().where(...).first()` and then check for `None` or not? Since username is supposed to be unique, you will either get one or none.

src/gateway/models.py
```diff
@@ -169,8 +169,6 @@ class User(BaseModel):
     id = AutoField()
     username = CharField()
```
khenderick, line 170: I think we forgot about the unique flag.

src/gateway/user_controller.py
```diff
@@ -130,92 +126,72 @@ def remove_user(self, username):
...
-                logger.debug("Removing expired token for user: [{}]".format(self._tokens[token][0]))
+                logger.debug('Removing expired token for user: "{}"'.format(self._tokens[token][0]))
                 self._tokens.pop(token, None)
 
         return ret
 
     def check_token(self, token):
         # type: (str) -> bool
-        """ Returns True if the token is valid, False if the token is invalid. """
+        """
```
khenderick, line 192: For single-line comments the compact notation is preferred. See style guide.

src/gateway/webservice.py
```diff
@@ -477,7 +468,10 @@ def get_usernames(self):
         """
         if not self.in_authorized_mode():
             raise cherrypy.HTTPError(401, "unauthorized")
-        return {'usernames': self._user_controller.get_usernames()}
+        users = self._user_controller.load_users()
+        usernames = [user.username for user in users]
+        usernames_str = ', '.join(usernames)
+        return {'usernames': usernames_str}
```
khenderick, line 474: If I'm not mistaken, that call returned a list of strings before, not a single string with comma separated values. Did you validate this behavior with the gateway web interface? I'd expect that the interface would show one records with this comma separated string and that deleting it would fail.

testing/unittests/gateway_tests/users_test.py
```diff
@@ -103,225 +101,296 @@ def test_terms(self):
...
+        self.assertEqual(1, len(users_in_controller))
+        self.assertEqual('om', users_in_controller[0].username)
+
+
+        fields = ['username', 'password', 'accepted_terms']
+
+        # create a new user to test with
+        user_dto = UserDTO(username='test')
+        user_dto.set_password('test')
         user_controller.save_users([(user_dto, fields)])
```
khenderick, line 272: If you create a test that validates loading users, it is best to use as few other functionality as possible. Here you also rely on the `save_users` implementation. If e.g. both save and load would actually not save to the ORM but for example keep the instances in memory, the test would pass.

Better for such a test is to directly insert ORM objects (you can find examples in other tests) and then test the retrieval methods. Similar, for a `save_users` test, best would then be to validate the ORM objects are actually in the DB by directly executing ORM queries.

On a side-note, I don't think the test name now matches the tested functionality, as the test doesn't test the `get_usernames` functionality anymore.

khenderick (changes requested):

khenderick (approved):

PR #436, opened 2020-09-15
Support for Gen3

khenderick:


src/master/core/slave_updater.py
```diff
@@ -101,6 +107,16 @@ def update(address, hex_filename, version, slave_communicator=INJECTED):  # type
                     logger.info('{0} - Already up-to-date. Skipping'.format(address))
                     return True
 
+                major_version = int(firmware_version.split('.')[0])
+                is_gen3 = major_version >= 6
+                gen3_firmware = module_type.endswith('3')
```
LnL7, line 112: Might be good to put these conditions in helper functions so it's not scattered around the code.

src/update.py
```diff
@@ -198,75 +243,73 @@ def check_master_communication():
...
     except Exception as exc:
-        logger.error('Updating {} firmware failed'.format(module))
+        logger.exception('Updating {} firmware failed'.format(module))
         return exc
 
 
-def update_gateway_os(tarball):
+def update_gateway_os(tarball, version):
     try:
+        if is_up_to_date('gateway_os', version):
```
LnL7, line 295: Doesn't this break re-applying an update, like the new update test?

khenderick: How do you mean? It will prevent the update from being applied if the version is already installed. But that's the requirement of this change.

src/master/core/slave_updater.py
khenderick: It's a bit awkward to have this slightly different code abstracted in some module higher up the directory tree to be able to share it. I've restructured some code so the "is-it-gen3-firmware" is now shared inside `modules_bootloader.py`

src/update.py
khenderick: After discussion on Slack, it indeed makes sense to (for now) always update the Gateway backend and Gateway OS.

LnL7 (approved):

PR #438, opened 2020-09-15
Ventilation ORM models

LnL7:


src/gateway/migrations/orm/012_add_ventilation.py
```diff
@@ -0,0 +1,50 @@
...
+
+    class Ventilation(BaseModel):
+        id = AutoField()
+        source = CharField()
+        source_id = IntegerField()
+        external_id = CharField()
+        name = CharField()
+        type = CharField()
+        vendor = CharField()
+        steps = IntegerField()
```
svanscho, line 43: As I'm reading this we should probably name this levels or n_levels or something.

PR #459, opened 2020-09-28
python3 requirements

LnL7:


pyproject.toml
```diff
@@ -5,34 +5,34 @@ description = "OpenMotics Gateway"
...
-peewee-migrate = "^1.1.6"
-psutil = "^5.6.7"
-pyserial = "^3.4.0"
-requests = "^2.18"
-simple-pid = "^0.2.4"
-sqlalchemy = "^1.3.12"
-typing = "^3.7.4"
-ujson = "^1.35"
-ws4py = "^0.5.1"
+apscheduler = "^3.6"
```
svanscho, line 10: this one I know needs to be very specific

svanscho (changes requested):

pyproject.toml
```diff
@@ -5,34 +5,34 @@ description = "OpenMotics Gateway"
 authors = []
 
 [tool.poetry.dependencies]
-python = "~2.7"
+python = "^2.7 || >=3.7"
```
svanscho, line 8: do we want 3.7 of something like 3.5?

svanscho (dismissed):

svanscho (approved):

pyproject.toml
LnL7: If we have the the choice I'd go straight to 3.7 or 3.8, depends on how we're going to ship python3.

svanscho: Debian out of the box supports 3.5 only I think. But I agree we should directly aim to get that updateable too.

PR #462, opened 2020-09-29
Feature/settings migration

ptuytsch:


src/gateway/config_controller.py
```diff
@@ -32,34 +33,15 @@
...
 
-    def __execute(self, *args, **kwargs):
-        with self.__lock:
-            try:
-                return self.__cursor.execute(*args, **kwargs)
-            except (sqlite3.OperationalError, sqlite3.InterfaceError):
-                time.sleep(randint(1, 20) / 10.0)
-                return self.__cursor.execute(*args, **kwargs)
-
     def __check_tables(self):
```
khenderick, line 41: Rename to e.g. `_insert_defaults` or similar.

src/gateway/config_controller.py
```diff
@@ -75,17 +57,41 @@ def __check_tables(self):
                 self.set(key, default_value)
 
     def get(self, key, fallback=None):
```
khenderick, line 59: It seems like these are now all static and the only reason to have an instance is to insert default values. Maybe an option could be to actually move these methods to the models itself (so you have a static `Config.get` etc)

You could then use the migrator to insert the defaults - more on that in another comment.

This would mean you can just remove the `ConfigController` completely.

src/gateway/dto/config.py
```diff
@@ -0,0 +1,39 @@
...
+    from typing import Optional, Any
+
+
+class ConfigDTO(BaseDTO):
+    def __init__(self, setting, data=None):
+        # type: (str, Optional[str]) -> None
+        self.setting = setting
+        self.data = data
+
+
```
khenderick, line 32: Spacing/formatting

src/gateway/dto/config.py
```diff
@@ -0,0 +1,39 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+settings DTO
```
khenderick, line 17: It's either config or setting.

src/gateway/metrics_controller.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, plugin_controller=INJECTED, metrics_collector=INJECTED, metri
         :param metrics_cache_controller: Metrics cache Controller
         :type metrics_cache_controller: gateway.metrics_caching.MetricsCacheController
         :param configuration_controller: Configuration Controller
-        :type configuration_controller: gateway.config.ConfigurationController
+        :type configuration_controller: gateway.config_controller.ConfigurationController
```
khenderick, line 60: Use mypy typings instead. You can then just remove all `:type` entries in this docstring.

src/gateway/migrations/config.py
```diff
@@ -0,0 +1,49 @@
...
+import os
+import logging
+import constants
+from gateway.migrations.base_migrator import BaseMigrator
+from gateway.models import Config
+
+logger = logging.getLogger('openmotics')
+
+
+class ConfigMigrator(BaseMigrator):
```
khenderick, line 25: In this class you could actually override the `migrate` method. At first, you call it's super (to make sure the actual migration is executed), but then you could basically add some code that inserts default values if they do not exist yet. This way if new setting/config entries are added, we can just add them here.

khenderick (changes requested):

src/gateway/migrations/config.py
```diff
@@ -0,0 +1,69 @@
...
+                setting = row[1]
+                config = Config.get_or_none(setting=setting)
+                if config is None:
+                    config = Config(
+                        setting=setting,
+                        data=row[2]
+                    )
+                    config.save()
+            os.rename(old_sqlite_db, '{0}.bak'.format(old_sqlite_db))
+            cls._insert_defaults()
```
khenderick, line 50: We should run this in an overridden `migrate` method, as the `_migrate` method is only executed once (based on a migration flag) inside the base class implementation. We'd need to run this `_insert_defaults` every time (so we can simply add records along the way without having to write new migrators every time).

khenderick (changes requested):

khenderick (approved):

PR #470, opened 2020-10-02
Few fixes and optimizations

khenderick:


src/gateway/hal/master_controller_core.py
```diff
@@ -472,20 +472,22 @@ def save_shutters(self, shutters):  # type: (List[Tuple[ShutterDTO, List[str]]])
             output_module = OutputConfiguration(shutter_dto.id * 2).module
             # Configure shutter
             shutter = ShutterMapper.dto_to_orm(shutter_dto, fields)
-            if shutter.timer_down is not None and shutter.timer_up is not None:
+            if shutter.timer_down not in [0, None] and shutter.timer_up not in [0, None]:
```
svanscho, line 475: perhaps we can override de __not__ operator?

khenderick: What comes out of `shutter.timer_down` and `shutter.timer_up` is simple integer. Nevertheless, it also means that `None` isn't a valid value and that should be `65535`. I was most likely confused with the DTO that could be optional.

svanscho (approved):

PR #474, opened 2020-10-06
split up commands

LnL7:


LnL7:
Replaced by #489 for now.

PR #478, opened 2020-10-08
Revert "version 2.17.0.9"

khenderick:
Reverts openmotics/gateway#477


khenderick:
Master is now on `2.16.8`, develop is on `2.17.0`, there's no need to bump any version now. After `develop` is merged to `master`, we can bump it to `2.17.1`.

PR #481, opened 2020-10-12
Added update status codes

svanscho:


src/update.py
```diff
@@ -489,10 +492,13 @@ def update(version, expected_md5):
         cmd('rm -v -rf {}/*'.format(update_dir), shell=True)
 
         if errors:
+            exitcode = EXIT_CODES['failed']
             logger.error('Exceptions:')
             for error in errors:
+                if isinstance(error, SystemExit):
+                    exitcode = error.code
```
LnL7, line 499: Using the exit code of the first failure probably makes more sense.

eg. something like

```
for error in errors:
    logger.error('- {0}'.format(error))
raise errors[0]
```

In combination with generic Exception handling in main with a default exit code instead of just SystemExit, which is probably a good idea anyway.

LnL7 (approved):

PR #484, opened 2020-10-15
master communication heartbeat

LnL7:


src/gateway/hal/master_controller_classic.py
```diff
@@ -1148,22 +1155,29 @@ def factory_reset(self):
...
-    def cold_reset(self):
-        # type: () -> None
+    def cold_reset(self, power_on=True):
+        # type: (bool) -> None
         """
         Perform a cold reset on the master. Turns the power off, waits 5 seconds and turns the power back on.
         """
-        self._master_online = False
         MasterClassicController._set_master_power(False)
         time.sleep(5)
```
khenderick, line 1164: Maybe it's better to move this sleep into the if-statement. Similar for the Core implementation.

src/gateway/watchdog.py
```diff
@@ -66,82 +72,64 @@ def stop(self):
...
-            return None
-        ratio = len([t for t in calls_last_x_minutes if t in calls_timedout]) / float(len(calls_last_x_minutes))
-        if ratio < 0.25:
-            # Less than 25% of the calls fail, let's assume everything is just "fine"
-            logger.warning('Noticed communication timeouts for \'{0}\', but there\'s only a failure ratio of {1:.2f}%.'.format(
-                name, ratio * 100
-            ))
-            return None
+        stats = controller.get_communication_statistics()
+        calls_timedout = [call for call in stats['calls_timedout']]
```
khenderick, line 102: You're not using the `start_time` variable anymore. This was to that if the watchdog is started it doesn't take previous statistics into account. If a test for example turns of the watchdog to e.g. validate certain behavior and tons of timeouts will be generated, and then the watchdog is restarted, it might immediately trigger on those previous timed out calls.

src/master/classic/master_communicator.py
```diff
@@ -102,14 +101,24 @@ def start(self):
...
             self.__read_thread.start()
 
     def stop(self):
         # type: () -> None
         self.__running = False
-        self.__read_thread.join()
+        if self.__read_thread:
+            self.__read_thread.join()
+            self.__read_thread = None
+        # self.__communication_stats = {'calls_succeeded': [],
```
khenderick, line 118: If not needed anymore, let's remove these comments.. On the other hand, does it make sense to reset the stats when the communicator restarts?

src/power/power_communicator.py
```diff
@@ -99,6 +102,39 @@ def get_communication_statistics(self):
...
+            # If there are no timeouts at all
+            return CommunicationStatus.SUCCESS
+
+        all_calls = sorted(calls_timedout + calls_succeeded)
+        calls_last_x_minutes = [t for t in all_calls if t > time.time() - 180]
+        ratio = len([t for t in calls_last_x_minutes if t in calls_timedout]) / float(len(calls_last_x_minutes))
+
+        if len(all_calls) <= 10:
+            # Not enough calls made to have a decent view on what's going on
+            logger.warning('Observed master communication failures, but not enough calls')
```
khenderick, line 121: `energy` instead of `master`

testing/cicd/tests/master_test.py
```diff
@@ -0,0 +1,61 @@
...
+    toolbox.health_check(timeout=120)
+    toolbox.dut.get('/get_status')
+
+
+@pytest.mark.slow
+def test_offline_recovery(toolbox, master_communication):
+    data = toolbox.dut.get('/get_status', success=False)
+    assert data['success'], data
+
+    toolbox.dut.get('/reset_master', {'power_on': False})
```
khenderick, line 54: What's the impact of this test to other tests if it leaves the master powered off?

khenderick (changes requested):
Few remarks. Also implement the functionality in the core communicator.

src/master/classic/master_communicator.py
LnL7: I did this initially to handle the condition from the comment above, but doing so significantly delays any action taken by the watchdog (with the current settings).

testing/cicd/tests/master_test.py
LnL7: The watchdog is expected to power on the master again.

src/master/core/core_communicator.py
```diff
@@ -89,6 +89,12 @@ def stop(self):
         self._read_thread.join()
 
     def get_communication_statistics(self):
+        self._communication_stats = {'calls_succeeded': [],
+                                     'calls_timedout': [],
+                                     'bytes_written': 0,
+                                     'bytes_read': 0}
+
+    def reset_communication_statistics(self):
```
khenderick, line 97: This seems the get, and above the reset.

khenderick (changes requested):

khenderick (approved):

khenderick (approved):

PR #485, opened 2020-10-19
gateway config events

LnL7:


src/gateway/base_controller.py
```diff
@@ -111,15 +114,21 @@ def _sync_orm(self):
...
                 except CommunicationTimedOutException as ex:
                     logger.error('ORM sync ({0}): Failed: {1}'.format(orm_model.__name__, ex))
                 except Exception:
                     logger.exception('ORM sync ({0}): Failed'.format(orm_model.__name__))
+
+            if self._sync_dirty:
+                type_name = orm_model.__name__.lower()
+                gateway_event = GatewayEvent(GatewayEvent.Types.CONFIG_CHANGE, {'type': type_name})
+                self._pubsub.publish_gateway_event(PubSub.GatewayTopics.CONFIG, gateway_event)
+                self._sync_dirty = False
```
svanscho, line 131: Don't we want this flag to be set to False when the cloud actively requests this? Same model as for the dirty EEPROM

svanscho (changes requested):

src/gateway/base_controller.py
```diff
@@ -111,15 +114,21 @@ def _sync_orm(self):
                             continue
                         id_ = dto.id
                         ids.append(id_)
-                        orm_model.get_or_create(number=id_)  # type: ignore
-                    orm_model.delete().where(orm_model.number.not_in(ids)).execute()  # type: ignore
+                        orm_model.get_or_create(number=id_)
```
svanscho, line 117: we should set the self._sync_dirty flag here in case something changed

svanscho (changes requested):

svanscho (approved):

PR #496, opened 2020-10-26
Added generic system information to the Heartbeat data

khenderick:


khenderick:
Back to the drawing board after review feedback

src/platform_utils.py
```diff
@@ -243,8 +243,10 @@ class Platform(object):
     class Type(object):
         CLASSIC = 'CLASSIC'
         CORE_PLUS = 'CORE_PLUS'
+        CORE = 'CORE'
 
-    Types = [Type.CLASSIC, Type.CORE_PLUS]
+    Types = [Type.CLASSIC, Type.CORE, Type.CORE_PLUS]
```
svanscho, line 248: perhaps you can do something like?
ClassicTypes = [Type.CLASSIC] 
CoreTypes = [Type.CORE, Type.CORE_PLUS]
Types = ClassicTypes + CoreTypes

svanscho (approved):

PR #499, opened 2020-10-28
add dummy platform

LnL7:


svanscho (approved):

src/platform_utils.py
```diff
@@ -240,13 +240,16 @@ class Platform(object):
...
+        DUMMY = 'DUMMY'
         CLASSIC = 'CLASSIC'
         CORE_PLUS = 'CORE_PLUS'
         CORE = 'CORE'
 
+    AnyTypes = [Type.DUMMY]
     ClassicTypes = [Type.CLASSIC]
     CoreTypes = [Type.CORE, Type.CORE_PLUS]
-    Types = ClassicTypes + CoreTypes
+    MasterTypes = ClassicTypes + CoreTypes
```
svanscho, line 251: These are not master types but rather hardware revisions. What will they be used for?

khenderick: There are three main parts here; hardware, system and platform. 

- Everything "hardware" in here is about the BB we're running on, so whether it's a BBB, BBGW, BBW, ... so we can figure out stuff like "what's the amount of memory", "how's the emmc doing", ...
- Everything in "system" is about the OS and things like that. So "what's the OS", "how do I start a service", ...
- Everything in "platform" is about what we're supposed to be; a classic gateway, a core, a coreplus, ...

So technically the platform will tell us about the master, but could gives us hints on what GPIO we should use, what serial ports can or should be used, ... So in this case, I don't think `MasterTypes` is a correct name. Maybe `PlatformTypes`, but we're already in the `Platform` class anyway.

LnL7: This was first used to decide whether the serial should be injected but that's not needed anymore since it's all optional now. If it's missing the injection fails now instead of the config parsing.

khenderick (approved):

PR #504, opened 2020-10-30
use constants in RS485 serial

LnL7:
This makes it a bit more clear what's going on exactly.

### DIFF
--- a/src/serial_utils.py
+++ b/src/serial_utils.py
@@ -56,6 +56,11 @@ def printable(data):
     return '{0}    {1}'.format(byte_notation, string_notation)
 
 
+TIOCSRS485 = 0x542F
+SER_RS485_ENABLED = 0b00000001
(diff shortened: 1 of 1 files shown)


src/serial_utils.py
```diff
@@ -65,8 +70,9 @@ def __init__(self, serial):
         self._serial = serial
         fileno = serial.fileno()
         if fileno is not None:
-            serial_rs485 = struct.pack('hhhhhhhh', 3, 0, 0, 0, 0, 0, 0, 0)
-            fcntl.ioctl(fileno, 0x542F, serial_rs485)
+            flags_rs485 = SER_RS485_ENABLED | SER_RS485_RTS_ON_SEND
```
svanscho, line 73: does this work for all platforms?

svanscho (approved):

PR #507, opened 2020-11-09
Allow longer pulse counter names

khenderick:


src/gateway/hal/mappers_classic/pulse_counter.py
```diff
@@ -41,7 +41,10 @@ def orm_to_dto(orm_object):  # type: (EepromModel) -> PulseCounterDTO
     def dto_to_orm(pulse_counter_dto, fields):  # type: (PulseCounterDTO, List[str]) -> EepromModel
         data = {'id': pulse_counter_dto.id}  # type: Dict[str, Any]
         if 'name' in fields:
-            data['name'] = pulse_counter_dto.name
+            name = pulse_counter_dto.name
+            if len(name) > 16:
+                name = '{0}~'.format(name[:15])
```
svanscho, line 46: I think you might want the last part as well. E.g.:
```
In [15]: for name in ["Output Bathroom 2", "Output Bathroom 1", "Output Bathroom 3"]:
    ...:     name = '{0}~'.format(name[:15])
    ...:     print(name)
    ...:
Output Bathroom~
Output Bathroom~
Output Bathroom~
```

svanscho (changes requested):

src/gateway/hal/mappers_classic/pulse_counter.py
svanscho: Perhaps something like: ```name='{0}~{1}'.format(name[:14], name[-1:])```
And perhaps we can put this in a toolbox somewhere to shorten all sorts of names in the eeprom?

khenderick: Yes, the EEPROM field is only 16 chars long so duplicates are inevitable. I think any not too expensive algorithm will have the same issue, but feel free to suggest one that has less chance for these duplicates.

svanscho (approved):

PR #517, opened 2020-12-02
Skip update for energy modules when the bus isn't configured

khenderick:


src/power_bootloader.py
```diff
@@ -299,6 +299,11 @@ def _get_from_ioc(power_store=INJECTED, power_communicator=INJECTED, power_seria
         return power_store, power_communicator, power_serial
 
     store, communicator, serial = _get_from_ioc()
+
```
svanscho, line 302: perhaps split this over multiple lines with detailed log line? saves another round trip of checking

if store is None:
    logger.info('Energy store not found')
if communicator is None:
    logger.info('Energy bus is disabled')
   ....

svanscho (changes requested):

src/power_bootloader.py
khenderick: As you can see here (https://github.com/openmotics/gateway/pull/517/files#diff-8a1a4eabd8ce905fc8d6a051b33944b4eba0b36fcd7d9732cdbb0185393cf77fR348) both will only be `None` when the bus is disabled.

svanscho: I can't clearly see the path, but I would assume there is a possibility the db is corrupt/missing while the serial is configured. I am not saying it's likely, but possible.
Sidenote: is it planned somewhere to also migrate this energy power store to the general gateway.db?

khenderick: There's a ticket for that

khenderick: Updated

svanscho (approved):

PR #522, opened 2020-12-03
Backup more files

khenderick:


src/update.py
```diff
@@ -336,7 +337,7 @@ def update_gateway_backend(tarball, date, version):
         if plugins:
             logger.info('Restoring plugins')
             for plugin in plugins:
-                cmd(['mv', '-v', plugin, os.path.join(python_dir, 'plugins')])
+                cmd(['cp', '-r', plugin, os.path.join(python_dir, 'plugins')])
```
svanscho, line 340:
```
man cp 
-v    Cause cp to be verbose, showing files as they are copied.
```

svanscho (changes requested):

src/update.py
svanscho: Perhaps good to add on the other commands too? I assume this will be present in the update logs then.

svanscho (changes requested):

src/update.py
LnL7: I recall something about `-v` not existing on angstrom.

svanscho: Sigh

khenderick: @svanscho does this imply the change is not required and the PR can be approved?

svanscho (approved):

PR #523, opened 2020-12-03
Fix/async pubsub

ptuytsch:


src/gateway/pubsub.py
```diff
@@ -16,11 +16,19 @@
 
 from collections import defaultdict
 import logging
+# Importing the Queue library as uniform naming convention
+import sys
+if sys.version_info.major == 3:
```
khenderick, line 21: You can import from six, see other places.

src/gateway/pubsub.py
```diff
@@ -47,12 +55,51 @@ def __init__(self):
         # type: () -> None
         self._gateway_topics = defaultdict(list)  # type: Dict[GATEWAY_TOPIC,List[Callable[[GatewayEvent],None]]]
         self._master_topics = defaultdict(list)  # type: Dict[MASTER_TOPIC,List[Callable[[MasterEvent],None]]]
+        self._master_events = Queue()  # type: Queue  # Queue[Tuple[str, MasterEvent]]
+        self._gateway_events = Queue()  # type: Queue  # Queue[Tuple[str, GatewayEvent]]
+        self.is_running = False
+        self._pub_thread = DaemonThread(name='Publisher loop',
+                                           target=self._publisher_loop,
```
khenderick, line 62: Proper code alignment here.

src/gateway/pubsub.py
```diff
@@ -47,12 +55,51 @@ def __init__(self):
...
+        # type: () -> None
+        self.is_running = True
+        self._pub_thread.start()
+
+    def stop(self):
+        # type: () -> None
+        self.is_running = False
+        self._pub_thread.stop()
+        self._master_events.join()
+        self._gateway_events.join()
```
khenderick, line 75: Better use `self._pub_thread.join()` here, see other comment.

src/gateway/pubsub.py
```diff
@@ -47,12 +55,51 @@ def __init__(self):
...
+
+    def _publish_all_events(self):
+        while not self._master_events.empty():
+            topic, master_event = self._master_events.get(block=True, timeout=5)
+            self._publish_master_event(topic, master_event)
+            self._master_events.task_done()
+        while not self._gateway_events.empty():
+            topic, gateway_event = self._gateway_events.get(block=True, timeout=5)
+            self._publish_gateway_event(topic, gateway_event)
+            self._gateway_events.task_done()
```
khenderick, line 92: It's better to use a `try / except Empty`. While here it's not strictly needed, it's good practice as it makes the code more thread-safe generally speaking.

Also, using a `timeout=5` below will cause the `.stop()` to be blocked up to 10 seconds in total, which will basically result in our service being killed instead of gracefully stopped. Also, since you will process this queue no matter what, it can even take more time. The thing is that when `.stop()` is called, we have to stop as quickly as possible.

Assuming you use a `try/except` here and a low timeout, the fastest way would be to add `None` to both queues in the `.stop()` method right after you set `self._running = False` You can then check for the value and check for None to abort the method. Something like:

```
def _publish_all_events(self):
  while True:
    try:
      event = self._master_events.get(block=True, timeout=0.25)
      if event is None:
        return
      self._publish_master_event(*event)  # Using * pours out the tuple over arguments
    except Empty:
      break
  while True:
    try:
      ...
```

khenderick (changes requested):

khenderick (approved):

PR #532, opened 2020-12-09
move configuration when removing a plugin

LnL7:
Plugin configuration can contain important data, renaming ensures that
re-installing or accidentally removing a plugin is less dangerous since
the config can still be recovered.

### DIFF
--- a/src/plugins/base.py
+++ b/src/plugins/base.py
@@ -366,7 +366,7 @@ class PluginController(object):
         # Remove the plugin configuration
         conf_file = '{0}/pi_{1}.conf'.format(self._plugin_config_path, name)
         if os.path.exists(conf_file):
-            os.remove(conf_file)
+            os.rename(conf_file, '{}.bak'.format(conf_file))
 
         # Finally remove database entry.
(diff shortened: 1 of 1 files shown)


khenderick:
I wonder whether it might also help to for example put the version in the backup filename. E.g. from `pi_Foo.conf` to `pi_Foo.1.2.3.conf` or similar. Then this could also be done when a plugin is updated; backup the current config file under the name of the previous value, and then update the plugin (which might cause a migration of the configuration file).

svanscho (approved):

src/plugins/base.py
```diff
@@ -366,7 +366,7 @@ def remove_plugin(self, name):
         # Remove the plugin configuration
         conf_file = '{0}/pi_{1}.conf'.format(self._plugin_config_path, name)
         if os.path.exists(conf_file):
-            os.remove(conf_file)
+            os.rename(conf_file, '{}.bak'.format(conf_file))
```
svanscho, line 369: will this work when the file already exists?

svanscho:
@LnL7 can we incorporate the changes and merge?

PR #539, opened 2020-12-14
Feature/pyinstaller build

ptuytsch:


src/openmotics_service.py
```diff
@@ -77,6 +80,12 @@ def setup_logger():
     handler.setFormatter(logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s"))
     logger.addHandler(handler)
 
+    if Hardware.get_board_type() == Hardware.BoardType.ESAFE:
```
khenderick, line 83: Better do this based on the buildroot OS as the location of logging is more of an OS thing and not directly related to used hardware.

src/openmotics_service.py
```diff
@@ -171,12 +180,17 @@ def start(
...
         # Execute data migration(s)
-        FeatureMigrator.migrate()
-        RoomsMigrator.migrate()
-        InputMigrator.migrate()
-        ScheduleMigrator.migrate()
-        UserMigrator.migrate()
-        ConfigMigrator.migrate()
+        config = ConfigParser()
+        config.read(constants.get_config_file())
+        # TODO: Make the master communication work before executing the migrations (needs eeprom use or other)
```
khenderick, line 185: If this code is skipped because the master isn't available at this setup, we should use the platform_utils's platform.

src/platform_utils.py
```diff
@@ -223,10 +234,14 @@ def handle_socket_exception(connection, exception, logger):
 
     @staticmethod
     def import_libs():
+        config = ConfigParser()
+        config.read(constants.get_config_file())
         operating_system = System.get_operating_system().get('ID')
-        if operating_system in (System.OS.ANGSTROM, System.OS.DEBIAN):
-            sys.path.insert(0, '/opt/openmotics/python-deps/lib/python2.7/dist-packages')
-            sys.path.insert(0, '/opt/openmotics/python-deps/lib/python2.7/site-packages')
+        if config.has_option('OpenMotics', 'build'):
```
khenderick, line 240: As discussed, rewrite this part to check whether it's running python2 or python3, and whether the respective folders are existing.

src/vpn_service.py
```diff
@@ -65,14 +64,26 @@ def setup_logger():
     handler.setFormatter(logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s'))
     logger.addHandler(handler)
 
+    if Hardware.get_board_type() == Hardware.BoardType.ESAFE:
```
khenderick, line 67: Same here, I think we should do this based on the OS buildroot.

src/vpn_service.py
```diff
@@ -65,14 +64,26 @@ def setup_logger():
...
+
 
 class VpnController(object):
     """ Contains methods to check the vpn status, start and stop the vpn. """
-
-    vpn_service = System.get_vpn_service()
-    start_cmd = 'systemctl start {0} > /dev/null'.format(vpn_service)
-    stop_cmd = 'systemctl stop {0} > /dev/null'.format(vpn_service)
-    check_cmd = 'systemctl is-active {0} > /dev/null'.format(vpn_service)
+    if Hardware.get_board_type() == Hardware.BoardType.ESAFE:
```
khenderick, line 76: Also base on OS buildroot

src/vpn_service.py
```diff
@@ -152,11 +166,17 @@ class Gateway(object):
 
     def __init__(self, host="127.0.0.1"):
         self._host = host
+        try:
```
khenderick, line 169: Use has_option to handle this instead of try/except.

gw-service/Makefile
```diff
@@ -0,0 +1,57 @@
...
+esafe-package:
+	echo "not implemented"
+
+${SPEC_FILE}:
+	. ./${VENV_FOLDER}/bin/activate; \
+	python --version; \
+	pyi-makespec \
+		--additional-hooks-dir=../hooks \
+		--onedir \
+		--add-data "../src/gateway/migrations:gateway/migrations" \
```
khenderick, line 32: Why is this folder different from other folders inside gateway?

khenderick (approved):

PR #554, opened 2021-01-19
Streamline logger

khenderick:
Fixes issue where the recently introduced `gateway` logging namespace would not be configured, so didn't log to anywhere.

### DIFF
--- a/src/logs.py
+++ b/src/logs.py
@@ -42,6 +42,10 @@ class Logs(object):
                             level=logging.INFO)
         openmotics_log_level = log_level
 
+        # Alter some system loggers
+        requests_logger = logging.getLogger('requests.packages.urllib3.connectionpool')
+        requests_logger.setLevel(logging.WARNING)
+
         # Prepare extra handlers
         openmotics_stream_handler = None
         if log_level != logging.INFO:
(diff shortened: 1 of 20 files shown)


src/power_bootloader.py
```diff
@@ -27,27 +27,22 @@
...
-
-    handler = logging.StreamHandler()
-    handler.setLevel(logging.INFO)
-    handler.setFormatter(logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s"))
-    logger.addHandler(handler)
+def extend_logger(_logger):
+    """ Extends the OpenMotics logger. """
+    _logger.setLevel(logging.DEBUG)
 
     handler = handlers.RotatingFileHandler(constants.get_update_log_location(), maxBytes=3 * 1024 ** 2, backupCount=2)
```
LnL7, line 42: This results in a new handler for each logger that points to the same file, that's not a problem?

src/logs.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Logs(object):
+
+    @staticmethod
+    def setup_logger(default_level=logging.INFO, extra_configuration=None):
+        """ Setup the OpenMotics logger. """
+
+        from platform_utils import System
+
+        handler = logging.StreamHandler()
```
LnL7, line 13: Since this mostly configures standard logging handlers `logging.basicConfig` could also be used to configure most of this in one go.

LnL7 (approved):

PR #555, opened 2021-01-19
Fix the plugin runtime on the esafe-device

ptuytsch:


src/openmotics_service.py
```diff
@@ -260,9 +262,25 @@ def stop(signum, frame):
...
 
 if __name__ == "__main__":
     setup_logger()
+
+    # First check if there are some arguments given, if so, check if it is for starting the plugin runtime
+    if len(sys.argv) > 1:
+        # delete the first argument to allign the other arguments properly, the first argument will be the name of this file
+        del sys.argv[0]
+        # check which file needs to be started up
+        if sys.argv[0] == 'runtime.py':
```
khenderick, line 279: As discussed, let's use a more user friendly keyword like `start_plugin` or `--start-plugin` something.

khenderick (changes requested):

khenderick (approved):

PR #559, opened 2021-01-28
Adding PluginWeb(Request/Response) objects to send between gateway an…

ptuytsch:
…d plugin


src/plugin_runtime/runtime.py
```diff
@@ -368,10 +380,16 @@ def _handle_request(self, method, args, kwargs):
             # Analog error message as the default CherryPy behavior
             return {'success': False, 'exception': 'Missing parameters: {0}'.format(', '.join(difference))}
         try:
-            return {'success': True, 'response': func(*args, **kwargs)}
+            if 'plugin_web_request' in kwargs:
```
khenderick, line 383: It's the best idea to check the decorated version here. If it's not version 2, the request object should never be given, as this might be some parameter currently used for something else. It's only when version 2 is explicitly used that we can be sure that this would behave as expected.

src/plugin_runtime/runtime.py
```diff
@@ -368,10 +380,16 @@ def _handle_request(self, method, args, kwargs):
...
+                kwargs['plugin_web_request'] = PluginWebRequest.from_serial(kwargs['plugin_web_request'])
+            func_return = func(*args, **kwargs)
+            if isinstance(func_return, PluginWebResponse):
+                func_return = func_return.serialize()
+            return {'success': True, 'response': func_return}
         except Exception as exception:
             return {'success': False, 'exception': str(exception), 'stacktrace': traceback.format_exc()}
 
+
```
khenderick, line 392: Extra whitespace not needed.

src/plugins/runner.py
```diff
@@ -37,23 +37,57 @@ def _cp_dispatch(self, vpath):
         # type: (List[str]) -> Any
         request = cherrypy.request
         response = cherrypy.response
-        method = vpath.pop()
+        path = '/'.join(vpath)
+        method = vpath[0]
+        # Clear vpath completely, The rest is not needed anymore
+        # The complete path is stored in the path variable
+        while len(vpath) > 0:
```
khenderick, line 44: Why is that needed? You can just let it be, right?

src/plugins/runner.py
```diff
@@ -37,23 +37,57 @@ def _cp_dispatch(self, vpath):
...
+                # Here it is read out when nessesary and put in the PluginWebRequest object at the correct place
+                if 'request_body' in kwargs:
+                    plugin_web_request.body = kwargs['request_body']
+                    del kwargs['request_body']
+                plugin_web_request.params = kwargs
+                kwargs = {'plugin_web_request': plugin_web_request.serialize()}
             contents = self.runner.request(method, args=args, kwargs=kwargs)
-            return contents.encode()
+            # See if the returned data fits the PluginWebResponse class
+            if PluginWebResponse.is_valid_serial_representation(contents):
```
khenderick, line 81: The if-else would not be needed if the runtime would always return a `PluginWebResponse`

src/plugins/runner.py
```diff
@@ -37,23 +37,57 @@ def _cp_dispatch(self, vpath):
...
+    def index(self, method, plugin_web_request=None, *args, **kwargs):
         try:
+            if plugin_web_request is not None:
+                # This has been placed under the 'request_body' in the webservice.py file
+                # Here it is read out when nessesary and put in the PluginWebRequest object at the correct place
+                if 'request_body' in kwargs:
+                    plugin_web_request.body = kwargs['request_body']
+                    del kwargs['request_body']
+                plugin_web_request.params = kwargs
+                kwargs = {'plugin_web_request': plugin_web_request.serialize()}
```
khenderick, line 78: This implies that a version 2 implementation on the plugin would only receive a single keyword argument `plugin_web_request`. Wouldn't it be more elegant if the get-parameters (if any) would still be passed as kwargs (if specified)? For a lot of the cases this works very easy and straightforward.

```
@om_expose(version=2)
def one(self, param_one, param_two):
   # .../one?param_one=1&param_two=2 still works

@om_expose
def two(self, param_one, param_two):
     # .../two?param_one=1&param_two=2 still works

@om_expose(version=2)
def three(self, param_one, param_two, plugin_web_request):
   # .../three?param_one=1&param_two=2 still works, but I can also access plugin_web_request['path'] or ...['body']

@om_expose(version=2)
def four(self, plugin_web_request):
   # .../four?param_one=1&param_two=2 still works, as is plugin_web_request['path'] or ...['body'], but I'll have to load e.g. plugin_web_request['params']['param_one'] myself
```

Not sure on the last one, I might be missing something, but it looks like GET params are not possible anymore when using the request body? Should still be possible right?

khenderick (changes requested):
A few more general remarks:
* Make sure to validate dependencies (e.g. base64) on older devices (use for example our Angstrom DUT) whether they are installed or not. If not, we'll have to provide a wheel.
* In the plugin tests, there's apart where "real" plugins are created, make sure to validate a few posibilities/versions there as well. Not only a static (de)serialization test, but also a full runner/runtime test.

khenderick (changes requested):

src/plugin_runtime/base.py
```diff
@@ -259,3 +263,172 @@ def _check_config(self, config, description):
...
+            else:
+                setattr(response, var, obj_dict[var])
+        return response
+
+    def __eq__(self, other):
+        if not isinstance(other, PluginWebResponse):
+            return False
+        vars_to_check = ['status_code', 'body', 'headers', 'path']
+        for var in vars_to_check:
+            if getattr(self, var) != getattr(other, var):
```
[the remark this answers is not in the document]
ptuytsch: The pluginWebBody is never used as a variable in the PluginWeb(Response/Request) objects. Only in the serialization and de-serialization process. But I've added it into the PluginWebBody. Better safe than sorry!

khenderick (approved):

PR #586, opened 2021-02-26
handle partial module discovery for basic cases

LnL7:
Before initialize could fail to discover the output module if it
was already discovered when some of the other modules are missing.

    Did not discover required modules: O=1. Raw log: []

### DIFF
--- a/testing/cicd/tests/toolbox.py
+++ b/testing/cicd/tests/toolbox.py
@@ -262,20 +262,26 @@ class Toolbox(object):
             expected_modules[hardware_type][module.mtype] += 1
         logger.info('Expected modules: {0}'.format(expected_modules))
 
-        try:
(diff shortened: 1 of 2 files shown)


testing/cicd/tests/toolbox.py
```diff
@@ -294,9 +300,9 @@ def initialize(self):
                 self.add_virtual_modules(module_amounts={mtype: extra_needed_amount})
 
         modules = self.count_modules('master')
-        logger.info('Current discovered modules: {0}'.format(modules))
+        logger.info('Virual modules: {0}'.format(modules))
```
svanscho, line 303: typo

svanscho (approved):

svanscho (approved):

PR #602, opened 2021-03-10
Fix/esafe update process

ptuytsch:
Fix the update process to also run on the eSafe devices with different code locations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,3 @@ venv/
 gw_py3_venv
 vpn-service/package/**
 gw-service/package/**
-src/log.py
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -16,3 +16,4 @@ venv/
 gw_py3_venv
 vpn-service/package/**
 gw-service/package/**
+src/log.py
```
khenderick, line 19: I don't think this should be ignored?

khenderick (changes requested):

.gitignore
ptuytsch: I removed the ignore line. But this was for me for a local debug log script since I never took the time to figure out the logs.py script.

khenderick (approved):

PR #603, opened 2021-03-11
Add online flag to ventilation status

ptuytsch:


src/gateway/dto/ventilation.py
```diff
@@ -75,13 +76,22 @@ class Mode:
...
+        # type: (int, str, Optional[int], Optional[float], Optional[float], Optional[float]) -> None
+        if timestamp is None:
+            # Set the default value for a timestamp as the current timestamp
+            timestamp = time.time()
         self.id = id
         self.mode = mode
         self.level = level
         self.timer = timer
         self.remaining_time = remaining_time
+        self.timestamp = timestamp
```
svanscho, line 89: nice shortcut can also be:
self.timestamp = timestamp or time.time()

svanscho, line 89: can we use something less generic than timestamp? e.g. last_seen or last_update?

src/gateway/ventilation_controller.py
```diff
@@ -57,13 +66,26 @@ def _publish_config(self):
...
-                      'timer': state_dto.timer}
+                      'timer': state_dto.timer,
+                      'is_connected': state_dto.is_connected}
         gateway_event = GatewayEvent(GatewayEvent.Types.VENTILATION_CHANGE, event_data)
         self._pubsub.publish_gateway_event(PubSub.GatewayTopics.STATE, gateway_event)
 
+    def _check_connected_timeout(self):
+        ventilation_items = list(self.last_ventilation_status.items())
+        for ventilation_id, ventilation_status_dto in ventilation_items:
+            delta = time.time() - ventilation_status_dto.timestamp
```
svanscho, line 81: this looks a bit duplicate with the is_connected in the DTO? It's a bit confusing to have a hardcoded value of 300 on one hand and a STATUS_TIMEOUT constant on the other side. Would either check the is_connected property here or remove the is_connected from the dto?

src/gateway/ventilation_controller.py
```diff
@@ -57,13 +66,26 @@ def _publish_config(self):
...
+            if delta > VentilationController.STATUS_TIMEOUT:
+                # timeout has passed, send a disconnect event
+                self._publish_state(ventilation_status_dto)
+                # Remove it from the last status since the status is disconnected, and it does not need to update again
+                del self.last_ventilation_status[ventilation_id]
```
svanscho, line 86: I would not expect a full delete of the status, but rather make it None for all relevant fields and keep the timestamp (or the connected flag, cfr comment above)

svanscho (changes requested):

src/gateway/ventilation_controller.py
```diff
@@ -57,13 +65,30 @@ def _publish_config(self):
...
                       'level': state_dto.level,
-                      'timer': state_dto.timer}
+                      'timer': state_dto.timer,
+                      'is_connected': state_dto.is_connected}
         gateway_event = GatewayEvent(GatewayEvent.Types.VENTILATION_CHANGE, event_data)
         self._pubsub.publish_gateway_event(PubSub.GatewayTopics.STATE, gateway_event)
 
+    def _check_connected_timeout(self):
+        ventilation_items = list(self.last_ventilation_status.items())
+        for ventilation_id, ventilation_status_dto in ventilation_items:
```
svanscho, line 79: is there a reason not to use for k,v in data.iteritems()?

svanscho (dismissed):

src/gateway/ventilation_controller.py
ptuytsch: There was one, but now there is none indeed. The list variable has been removed.

svanscho (approved):

PR #614, opened 2021-03-19
Scheduling bug fixed - unit test for the bug written - some cleanup

Shinigami-shinto:


testing/unittests/gateway_tests/scheduling_test.py
```diff
@@ -143,6 +143,7 @@ def test_base_validation(self):
             controller._validate(schedule)
 
     def test_group_action(self):
+        fakesleep.monkey_restore()
```
LnL7, line 146: Given that the tests have problems with thread coordination it's probably better to get split the test class. One for the tests that use fakesleep and one that doesn't so this doesn't need to be added to each case separately.

src/gateway/scheduling.py
```diff
@@ -140,57 +154,74 @@ def _process(self):
...
-            logger.error('Got error while executing schedule: {0}'.format(ex))
-            schedule_dto.last_executed = time.time()
-        finally:
-            schedule_dto.running = False
-            schedule_dto.next_execution = SchedulingController._get_next_execution(schedule_dto)
+    def _execute_schedule(self, schedules_to_execute, schedule):
+        # type: (List[ScheduleDTO], Schedule) -> None
+        for schedule_dto in schedules_to_execute:
+            if schedule_dto.running:
+                return
```
LnL7, line 194: Since this now handles multiple schedules this should only break out of the current loop.

```suggestion
                continue
```

Shinigami-shinto: nice catch :)

LnL7 (approved):

PR #615, opened 2021-03-22
Bugfix/ventilation events not send

ptuytsch:


testing/unittests/gateway_tests/ventilation_controller_test.py
```diff
@@ -331,20 +331,3 @@ def callback(event):
             self.pubsub._publish_all_events()
             self.assertEqual(2, len(events))
             self.assertEqual(1, len(self.controller.last_ventilation_status))
-
```
ptuytsch, line 334: These test statements are no longer useful since the gateway will not check periodically check the is_connected flag. It will just continuously send an event of the ventilation box.

svanscho (approved):

PR #617, opened 2021-03-23
add PyJWT and cryptography

LnL7:
These will be needed for signing RS256 jwt tokens.


svanscho:
Perhaps it would make sense to add this into the cloud-sdk library and make 2 versions to use as an import for debian/angstrom?

svanscho (approved):

LnL7:
These wheels are not compatible with stock gateways.

PR #619, opened 2021-03-24
fix wait condition in scheduling and filter past executions

LnL7:
Waiting with `now - next_start` would always continue immediately since
the next schedule time is supposed to be larger.  Also handle
potentially triggering old schedules and calculating an outdated
time window to batch schedules.


Shinigami-shinto:
Dikke fout in geslopen :facepalm:

Shinigami-shinto:
Dikke fout in geslopen :facepalm:
Goed dat je dat hebt gespot, held :)

khenderick (approved):

PR #623, opened 2021-03-26
Feature/esafe models

ptuytsch:
Addition of the models required for eSafe integration


src/gateway/api/serializers/apartment.py
```diff
@@ -0,0 +1,62 @@
...
+    @staticmethod
+    def deserialize(api_data):
+        # type: (Dict[str,Any]) -> Tuple[ApartmentDTO, List[str]]
+        loaded_fields = []
+        apartment_id = None
+        if 'id' in api_data:
+            loaded_fields.append('id')
+            apartment_id = api_data['id']
+        name = ''
+        if 'name' in api_data:
```
khenderick, line 50: This deserialisation be simplified using `SerializerToolbox.deserialize`. See for example in the `OutputSerializer`.

src/gateway/dto/user.py
```diff
@@ -53,17 +81,25 @@ def clear_password(self):
...
+        Sets the hashed password field of the UserDTO object, this way no clear-text passwords are used.
         """
         if password == '':
             raise ValueError("Password cannot be empty")
 
         self.hashed_password = UserDTO._hash_password(password)
 
     def __str__(self):
-        return "UserDTO: {}".format(vars(self))
+        return "UserDTO: first_name: {}, last_name: {}, role: {}".format(self.first_name, self.last_name, self.role)
```
khenderick, line 92: Maybe better remove this method so it uses the methods on the `BaseDTO` so e.g. logs will be more consistent over different objects

src/gateway/migrations/orm/017_add_esafe_models.py
```diff
@@ -0,0 +1,121 @@
...
+    #     class Meta:
+    #         table_name = '_user_old'
+    #     id = AutoField()
+    #     username = CharField(unique=True)
+    #     password = CharField()
+    #     accepted_terms = IntegerField(default=0)
+
+    # USER TARGET MODEL:
+    # ---------------------
+    class User(BaseModel):
```
khenderick, line 56: If I recall correctly you should define the models as they are before the migration, and then add the fields using the migrator. See for example `010_change_modules.py` or `006_change_inputs.py`. Then I'd think the migration would be easier.

testing/unittests/gateway_tests/users_test.py
```diff
@@ -73,10 +99,18 @@ def test_empty(self):
...
         user_dto.set_password("pass")
 
-        # verfify that the cloud user can login
+        print('printing users')
+        fields = ['first_name', 'last_name', 'password', 'role', 'pin_code']
+        for u in User.select():
+            user_str = ', '.join(['{}: {}'.format(x, getattr(u, x)) for x in fields])
+            print('user: {}'.format(user_str))
+
+        print(user_dto)
```
khenderick, line 111: Old print statement

khenderick (changes requested):

src/gateway/migrations/orm/018_move_user_data.py
```diff
@@ -0,0 +1,89 @@
...
+        user.role = User.UserRoles.ADMIN  # Set all existing users as admin users
+        user.save()
+
+    # Change the first name back to non null values
+    migrator.change_fields(User, first_name=CharField(null=False))
+    # Change the pin code field after the fact since it will not allow to be set unique when adding the column
+    # The pin_code is by definition unique since it is a copy of username for the existing users,
+    # (making the pin code unusable for the most part) which is also unique.
+    migrator.change_fields(User, pin_code=CharField(null=False, unique=True))
+
```
khenderick, line 85: Let's also remove the `username_old` field if we don't want to use that anymore.

khenderick (changes requested):

khenderick (approved):

khenderick (approved):

PR #626, opened 2021-03-31
Feature/v1 api

ptuytsch:


src/gateway/api/serializers/user.py
```diff
@@ -50,8 +52,10 @@ def serialize(dto_object, fields=None):
     def deserialize(api_data):
         # type: (Dict[str,Any]) -> UserDTO
         user_id = api_data['id'] if 'id' in api_data else None
-        user_dto = UserDTO(user_id)
-        for field in ['first_name', 'last_name', 'role', 'pin_code', 'accepted_terms']:
+        first_name = api_data['first_name'] if 'first_name' in api_data and api_data['first_name'] is not None else ''
+        last_name = api_data['last_name'] if 'last_name' in api_data and api_data['last_name'] is not None else ''
+        user_dto = UserDTO(user_id, first_name=first_name, last_name=last_name)
```
khenderick, line 57: This means that `first_name` and `last_name` will always be in the loaded_fields, and thus probably updated, even if the field wasn't passed to the API. This could cause a first name to be emptied while the API client would not expect. If both are mandatory, I think it would be more correct to validate whether the fields are filled in the webservice implementation and just use it here.

src/gateway/user_controller.py
```diff
@@ -91,98 +111,82 @@ def load_users(self):
...
-        if token is None or token not in self._tokens:
-            return False
-        else:
-            timed_out = self._tokens[token][1] >= time.time()
-            return timed_out
+        result = self.authentication_controller.check_token(token)
+        return result is not None
+
+    def generate_new_pin_code(self):
+        # loads the users in the self.users_cache
```
khenderick, line 166: Is this comment accurate?

src/gateway/webservice_v1.py
```diff
@@ -0,0 +1,441 @@
...
+                                      controller=self, action='get_user',
+                                      conditions={'method': ['GET']})
+        self.route_dispatcher.connect('get_available_code', '/available_code',
+                                      controller=self, action='get_available_code',
+                                      conditions={'method': ['GET']})
+        # --- POST ---
+        self.route_dispatcher.connect('post_user', '',
+                                      controller=self, action='post_user',
+                                      conditions={'method': ['POST']})
+        self.route_dispatcher.connect('post_activate_user', '/activate/:user_id',
```
khenderick, line 222: Wouldn't `/:user_id/activate` be more logical?

src/gateway/api/serializers/user.py
ptuytsch: Nice catch, The fields are not required in the api, and the logic has been altered. Also, some tests have been added for the user serializer

khenderick (approved):

PR #627, opened 2021-04-01
Simplify DTOs

khenderick:


src/gateway/api/serializers/rtd10.py
```diff
@@ -41,22 +41,17 @@ def serialize(global_rtd10_dto, fields):  # type: (GlobalRTD10DTO, Optional[List
...
-        loaded_fields = []
+    def deserialize(api_data):  # type: (Dict) -> GlobalRTD10DTO
         heating_values = {}
         cooling_values = {}
         for temperature in GlobalRTD10DTO.TEMPERATURES:
             field = 'output_value_heating_{0}'.format(GlobalRTD10Serializer._temp_to_str(temperature))
-            if field in api_data:
-                loaded_fields.append(field)
-                heating_values[temperature] = api_data[field]
+            heating_values[temperature] = api_data[field]
```
ptuytsch, line 49: Why removing the `if field in api_data` check? Is it guaranteed to always have the field when deserialized?

src/gateway/api/serializers/rtd10.py
```diff
@@ -41,22 +41,17 @@ def serialize(global_rtd10_dto, fields):  # type: (GlobalRTD10DTO, Optional[List
...
             field = 'output_value_cooling_{0}'.format(GlobalRTD10Serializer._temp_to_str(temperature))
-            if field in api_data:
-                loaded_fields.append(field)
-                cooling_values[temperature] = api_data[field]
+            cooling_values[temperature] = api_data[field]
```
ptuytsch, line 51: Same here for the `if field in api_data` check

src/gateway/hal/mappers_classic/rtd10.py
```diff
@@ -44,15 +44,13 @@ def orm_to_dto(orm_object):  # type: (EepromModel) -> GlobalRTD10DTO
...
 
     @staticmethod
-    def dto_to_orm(global_rtd10_dto, fields):  # type: (GlobalRTD10DTO, List[str]) -> EepromModel
+    def dto_to_orm(global_rtd10_dto):  # type: (GlobalRTD10DTO) -> EepromModel
         data = {}
         for temperature in GlobalRTD10DTO.TEMPERATURES:
             field = 'output_value_heating_{0}'.format(GlobalRTD10Mapper._temp_to_str(temperature))
-            if field in fields:
-                data[field] = global_rtd10_dto.heating_values[temperature]
+            data[field] = global_rtd10_dto.heating_values[temperature]
```
ptuytsch, line 51: The same `if field in fields` check is not needed?

src/gateway/room_controller.py
```diff
@@ -59,16 +59,16 @@ def load_rooms(self):  # type: () -> List[RoomDTO]
...
+            for room_dto in rooms:
                 if room_dto.in_use:
-                    room = RoomMapper.dto_to_orm(room_dto, fields)
-                    if 'floor' in fields:
+                    room = RoomMapper.dto_to_orm(room_dto)
+                    if 'floor' in room_dto.loaded_fields:
                         floor = None
                         if room_dto.floor is not None:
-                            floor = FloorMapper.dto_to_orm(room_dto.floor, ['id'])
+                            floor = FloorMapper.dto_to_orm(room_dto.floor)
```
ptuytsch, line 71: In this case, could it not be that you save more than needed? In the "old" version you only explicitly tell the mapper to only map the `id` field. Are you sure that only the id will be loaded in the floor DTO? Or does this not matter in this case?

ptuytsch: Disregard this comment when save rooms is only called to create new rooms and not to override existing rooms.

ptuytsch (approved):

src/gateway/api/serializers/rtd10.py
khenderick: This is unsure to be honest; the RTD10 is an obsolete integration and will/should be replaced by a plugin instead. It's not integrated in our user interfaces and is typically configured through the CLI interface, but since at some point in history it was added as an API we kinda maintain it. However, as the code would be quite a bit more complex to keep the API idential, the introduction of this small backwards incompatibility is a risk we can take.

We've noticed that only one (cloud connected) customer uses this feature, and this customer is not using this API.

src/gateway/room_controller.py
khenderick: We can possibly save more, but I don't think that would be an issue to be honest as in that case the data would be explicitly added to the DTO, so I think it's fine to save it as well.

PR #628, opened 2021-04-02
Added support for ShutterGroups

khenderick:


src/gateway/hal/master_controller_core.py
```diff
@@ -594,7 +603,7 @@ def load_shutter_groups(self):  # type: () -> List[ShutterGroupDTO]
         return shutter_groups
 
     def save_shutter_groups(self, shutter_groups):  # type: (List[ShutterGroupDTO]) -> None
-        pass  # TODO: Implement when/if ShutterGroups get actual properties
+        pass  # ShutterGroups have no properties on the Core
```
svanscho, line 606: shouldn't we raise instead of silently ignore this one?

khenderick: Currently that's what we do with all functionality that is not yet implemented.

svanscho: Isn't it time to switch to raising errors and have the portal handle this (needed for other parts too I assume)? Or you think it's too early for that? @iOlivier

svanscho: As a side remark, this is not an API which is not _yet_ implemented. It's just not supported, at any time.

svanscho (approved):

PR #660, opened 2021-04-27
fix create user api, not all fields where filled in

ptuytsch:


src/gateway/models.py
```diff
@@ -549,10 +549,11 @@ class UserLanguages(object):
         FR = 'Francais'
 
     id = AutoField()
-    first_name = CharField(null=False)
-    last_name = CharField(null=False, default='')
+    _username = CharField(null=False, unique=True)
```
LnL7, line 552: Why not keep the username field as before and move the python logic of the property to the defaults during creation.

src/gateway/dto/user.py
```diff
@@ -28,13 +28,14 @@
...
         self.id = id  # type: Optional[int]
-        if username != '':
-            self._username = username  # type: Optional[str]
-        else:
-            self._username = None
+        self.username = username  # type: str
+        # if there is no username, but one can be created from the first and last name, create it as well
+        if username is None and (first_name != '' or last_name != ''):
+            self.username = '{}.{}'.format(first_name.replace(' ', '.').lower(), last_name.replace(' ', '.').lower())  # type: str
+            self.loaded_fields.append('username')  # Append username to the loaded fields
```
LnL7, line 38: I don't think this is necessary, the property setter will do this automatically.

src/gateway/dto/user.py
```diff
@@ -97,15 +74,14 @@ def set_password(self, password):
             raise ValueError("Password cannot be empty")
 
         self.hashed_password = UserDTO._hash_password(password)
-        self._loaded_fields.add('password')
-        if '_hashed_password' in self._loaded_fields:
-            self._loaded_fields.remove('_hashed_password')
+        self.loaded_fields.append('hashed_password')
```
LnL7, line 77: Same as above.

src/gateway/user_controller.py
```diff
@@ -145,8 +143,7 @@ def login(self, user_dto, accept_terms=False, timeout=None):
             timeout = self._token_timeout
 
         user_orm = User.select().where(
-            (User.first_name == user_dto.first_name.lower()) &
-            (User.last_name == user_dto.last_name.lower()) &
+            (User.username == user_dto.username.lower()) &
```
LnL7, line 146: Does this filtering work as expected? I've noticed issues with `&` combined queries during testing.

LnL7 (approved):

LnL7 (approved):

PR #662, opened 2021-04-27
Fixed unwanted filtering of input_status v2 decorated method

svanscho:
Input status events with version 2 were filtered.

Refactored some methods along the way to be in line with other code.

### DIFF
--- a/src/plugins/runner.py
+++ b/src/plugins/runner.py
@@ -246,7 +246,7 @@ class PluginRunner(object):
 
     def process_input_status(self, input_event):
         event_json = input_event.serialize()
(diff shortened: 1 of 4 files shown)


src/plugins/runner.py
```diff
@@ -246,7 +246,7 @@ def stop(self):
 
     def process_input_status(self, input_event):
         event_json = input_event.serialize()
-        self._do_async(action='input_status', payload={'event': event_json}, should_filter=True)
+        self._do_async(action='input_status', payload={'event': event_json}, should_filter=False)
```
svanscho, line 249: @LnL7 this is the culprit

LnL7 (approved):

LnL7 (approved):

PR #673, opened 2021-05-04
Code cleanup

khenderick:


src/gateway/input_controller.py
```diff
@@ -78,6 +89,28 @@ def save_inputs(self, inputs):  # type: (List[InputDTO]) -> None
             inputs_to_save.append(input_dto)
         self._master_controller.save_inputs(inputs_to_save)
 
+    def get_input_status(self):  # type: () -> List[Dict[str, Any]]
+        """
+        Get a list containing the status of the Inputs.
+        :returns: A list is a dicts containing the following keys: id, status.
+        """
+        # TODO: Return a List[DTO]
+        return [{'id': input_port['id'], 'status': input_port['status']}
```
LnL7, line 98: Would be nice to also rework this to cache the input status in the controller similar to outputs, instead of calling straight to the master.

src/gateway/webservice.py
```diff
@@ -1619,54 +1615,36 @@ def set_group_action_configurations(self, config):  # type: (List[Dict[Any, Any]
...
-        :type id: int
-        :param fields: The field of the scheduled_action_configuration to get. (None gets all fields)
-        :type fields: list
-        :returns: 'config': scheduled_action_configuration dict: contains 'id' (Id), 'action' (Actions[1]), 'day' (Byte), 'hour' (Byte), 'minute' (Byte)
-        :rtype: dict
-        """
-        return {'config': self._gateway_api.get_scheduled_action_configuration(id, fields)}
+        # type: (int, Optional[List[str]]) -> Dict[str, Any]
+        """ Get a specific scheduled_action_configuration defined by its id. """
+        return {'config': LegacyScheduleSerializer.serialize(schedule_dto=self._scheduling_controller.load_scheduled_action(id),
```
LnL7, line 1624: Since these calls where changed a bit a webservice test for the scheduling and dimmers would be nice.

src/gateway/system_controller.py
```diff
@@ -0,0 +1,267 @@
...
+            logger.info('Restarting for factory reset...')
+            System.restart_service('openmotics')
+
+        Timer(2, _restart).start()
+        if can:
+            return {'factory_reset_full': 'pending'}
+        return {'factory_reset': 'pending'}
+
+    def get_master_backup(self):
+        return self._master_controller.get_backup()
```
khenderick, line 257: Move to the `ModuleController` and use that one from here.

src/gateway/input_controller.py
khenderick: I've updated the TODO to mention this, but let's do this in a separate PR

src/gateway/webservice.py
khenderick: Tests added (and bug caught)

LnL7 (approved):

PR #674, opened 2021-05-06
OM2185: Reverse plugin logs

svanscho:
image

@iOlivier @khenderick it seems we would need portal changes too as it's not auto refreshing for some reason.


khenderick (approved):
Correct, at this moment, the APIs always return the full log buffer. So the portal parses the timestamps, and then matches these to figure out the new lines and adds them to the view.

While this change looks technically OK (I'll approve), I'm a bit confused why we'd reverse the plugin logs on the GW as logs are always from old to new imo. If we want to display them in reverse, I'd think this would be a frontend-only change (which wouldn't be too hard, sinds it has to parse them anyway).

svanscho:
@iOlivier I am not merging this PR. Better option would be to probably auto-scroll on the portal?

PR #681, opened 2021-05-12
Input Controller status thread

svanscho:
Input status is null on the cloud as there are no periodic updates sent from the gateway.


src/gateway/hal/master_controller_classic.py
```diff
@@ -401,13 +393,31 @@ def get_input_module_type(self, input_module_id):
         o = self._eeprom_controller.read(eeprom_models.InputConfiguration, input_module_id * 8, ['module_type'])
         return o.module_type
 
+    @communication_enabled
     def get_inputs_with_status(self):
```
LnL7, line 397: Is there a reason to keep the dict responses?

src/gateway/input_controller.py
```diff
@@ -114,3 +181,62 @@ def get_input_module_type(self, input_module_id):
     def load_inputs_event_enabled():
         return {input_['number']: input_['event_enabled']
                 for input_ in Input.select().dicts()}
+
+
+class InputStateCache(object):
+    def __init__(self):
+        self._cache = {}  # type: Dict[int,InputDTO]
```
LnL7, line 188: It looks like everything here only uses the input status.

pyproject.toml
```diff
@@ -33,6 +33,7 @@ mock = "^3.0"
 pytest = "^4.6"
 xmlrunner = "^1.7"
 pytest-cov = "^2.10"
+freezegun = "^0.3.15"
```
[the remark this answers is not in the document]
svanscho: ok

src/gateway/hal/master_controller_classic.py
svanscho: we also have the load_input_status, this is where we return InputStatusDTO objects already.

src/gateway/input_controller.py
svanscho: this is inline with the output controller, and can be changed if needed. But would advise to use the same approach in both.

src/gateway/input_controller.py
```diff
@@ -39,15 +49,63 @@ class InputController(BaseController):
...
+                      'location': {'room_id': Toolbox.denonify(input_dto.room, 255)}}
+        gateway_event = GatewayEvent(event_type=GatewayEvent.Types.INPUT_CHANGE, data=event_data)
+        self._pubsub.publish_gateway_event(PubSub.GatewayTopics.STATE, gateway_event)
+
     def _handle_master_event(self, master_event):
         # type: (MasterEvent) -> None
+        super(InputController, self)._handle_master_event(master_event)
+        if master_event.type == MasterEvent.Types.MODULE_DISCOVERY:
+            if self._sync_state_thread:
+                self._sync_state_thread.request_single_run()
```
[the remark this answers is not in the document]
svanscho: This is the same approach as for the output controller. To avoid regressions, I took the same approach.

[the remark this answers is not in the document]
LnL7: the EEPORM master event is the point that the base controller hooks in which would be the correct event type, but even in that case the actual orm sync for inputs won't have completed yet so nothing will be updated.

https://github.com/openmotics/gateway/pull/681/files#diff-8e2c5e2fd1a305803ac43be4cfc96ff1bc4f18f4e446a94e63030c8d68befb50R232-R234

src/gateway/hal/master_controller_classic.py
svanscho: changed

LnL7 (approved):

PR #682, opened 2021-05-12
Update from safety bit to safety counter

khenderick:


ptuytsch (commented):
Is there any notion of the version of the ucan firmware? Since I don't see how you can clear the safety_flag on the older firmware? (see the firmware repo)

svanscho (approved):

khenderick:
> Is there any notion of the version of the ucan firmware? Since I don't see how you can clear the safety_flag on the older firmware? (see the firmware repo)

The older uCANs do not have a bootloader at all, and there should never be uCANs in the field with a bootloader that does not support this feature.

PR #685, opened 2021-05-14
OM-2169 Feature/v1 system api

ptuytsch:


src/gateway/system_config_controller.py
```diff
@@ -0,0 +1,184 @@
...
+        # type: () -> SystemRFIDSectorBlockConfigDTO
+        conf = cls._get_config_values(['rfid_sector_block'])
+        return SystemRFIDSectorBlockConfigDTO(**conf)
+
+    @classmethod
+    def save_rfid_sector_block_config(cls, config_dto):
+        # type: (SystemRFIDSectorBlockConfigDTO) -> None
+        cls._save_config_values(config_dto)
+
+    @classmethod
```
LnL7, line 138: Maybe this belongs a bit more in the system controller since it's not really just configuration.

src/constants.py
```diff
@@ -208,3 +208,7 @@ def get_runit_service_folder():
 def get_renson_main_config_file():
     # type: () -> str
     return '/data/app_data/main.config'
+
+def get_esafe_touchscreen_calibration_file():
+    # type: () -> str
+    return '/etc/pointercal'
```
LnL7, line 214: Isn't `/etc` readonly? This seems more like state that belongs under the application data.

src/gateway/system_config_controller.py
```diff
@@ -0,0 +1,184 @@
...
+@Singleton
+class SystemConfigController(object):
+
+    def __init__(self):
+        pass
+
+    ESAFE_CONFIG_PREFIX = 'ESAFE_'
+    DEFAULT_CONFIG = {
+        # 'db_version': '006',
+        'ESAFE_doorbell_enabled': True,
```
LnL7, line 43: Config entries in the database are stored as json encoded values so values that are always updated together could be placed under a shared key.

LnL7 (commented):
General remark, how much of these setting DTOs will be used actively by the backend? A more generic api like the current `/get_setting` means that to add a new setting is much simpler. Especially for simple enable flags there doesn't seem to be much added benefit to have separate api endpoints except for validation.

src/constants.py
ptuytsch: Good point, it seems that the eSafe calibration api does not work on the original backend as well... The file can indeed not be written. I suspect that the API is used when the device is setup, and when it leaves the factory, the device is set in readonly mode

src/gateway/system_config_controller.py
ptuytsch: This is indeed possible, but this makes it a lot harder to check if the new value is correct and does not add some new values to the config of the common shared key. I've made a prototype of how this could work, but this seems not to be willing to let it update itself easily. If possible, could we discuss this offline?

LnL7 (approved):

PR #699, opened 2021-05-21
Add generic equal function to the baseDTO class

ptuytsch:


LnL7 (approved):

khenderick:
A small remark, wouldn't it be a good idea to also compare `self.loaded_fields` with `other.loaded_fields`? Otherwise any DTO with for example only the `id` loaded, will always be equal to any other instance with the same ID and whatever values are in other (loaded) fields.

PR #704, opened 2021-05-25
input/output status sync

LnL7:


src/gateway/base_controller.py
```diff
@@ -80,6 +80,10 @@ def stop(self):
         if self._sync_orm_thread is not None:
             self._sync_orm_thread.stop()
 
+    def request_sync_state(self):
```
svanscho, line 83: perhaps we should override this for the sensors too?

LnL7: There's currently no state syncing like this in the controller since classic/core+ behave quite differently in terms of master events.

svanscho (approved):

PR #705, opened 2021-05-25
Feature/generic eq dto function

ptuytsch:


src/gateway/dto/base.py
```diff
@@ -44,9 +44,14 @@ def __setattr__(self, key, value):
 
     def __eq__(self, other):
         # type: (Any) -> bool
+        # This only works with the generic DTO which uses the 'capture_fields' decorator
+        field_names = getattr(self.__class__.__init__, 'field_names', None)
```
LnL7, line 48: If `__eq__` depends on the decorator we should move that logic to the class itself using a metaclass so it can't break in unexpected ways. Alternatively comparing the loaded fields should also fix the issue.

```python
if set(self.loaded_fields) != set(other.loaded_fields):
    return False
```

LnL7 (dismissed):

LnL7 (dismissed):

LnL7 (approved):

PR #710, opened 2021-06-07
Authentication V1 API

ptuytsch:


src/cloud/events.py
```diff
@@ -82,7 +82,7 @@ def _send_events_loop(self):
             if not self._batch_send_events():
                 raise DaemonThreadWait
         except APIException as ex:
-            logger.error('Error sending events to the cloud %', ex)
+            logger.error('Error sending events to the cloud {}'.format(ex))
```
LnL7, line 85: Using percent formatting logging is generally best practice https://docs.python.org/2.7/library/logging.html#module-level-functions, but just % isn't a valid formatter.

```suggestion
            logger.error('Error sending events to the cloud %s', ex)
```

src/gateway/authentication_controller.py
```diff
@@ -88,6 +90,27 @@ def login(self, user_dto, accept_terms=False, timeout=None):
...
+
+        if user_orm is None:
+            return False, UserEnums.AuthenticationErrors.INVALID_CREDENTIALS
+
+        user_dto = UserMapper.orm_to_dto(user_orm)
+        return self.login(user_dto, accept_terms=accept_terms, timeout=timeout)
+
+    def login_with_rfid_tag(self, rfid_tag_string, accept_terms=False, timeout=None):
+        # type: (str, bool, Optional[float]) -> Tuple[bool, Union[str, AuthenticationToken]]
+        """  Login a user given a UserDTO """
```
LnL7, line 108:
```suggestion
        """  Login a user using an authorized RFID tag """
```

src/gateway/webservice_v1.py
```diff
@@ -849,6 +850,43 @@ def delete_rfid(self, rfid_id, token=None):
...
+        # --- POST ---
+        self.route_dispatcher.connect('authenticate', '/authenticate',
+                                      controller=self, action='authenticate',
+                                      conditions={'method': ['POST']})
+        self.route_dispatcher.connect('deauthenticate', '/deauthenticate',
+                                      controller=self, action='deauthenticate',
+                                      conditions={'method': ['POST']})
+
+    @openmotics_api_v1(auth=False, expect_body_type='JSON')
+    def authenticate(self, request_body):
```
LnL7, line 870: Is the current esafe backend api also overloaded? Splitting up the endpoints for each type might be a good idea.

src/gateway/webservice_v1.py
```diff
@@ -849,6 +850,43 @@ def delete_rfid(self, rfid_id, token=None):
...
+        if 'code' in request_body:
+            success, data = self._user_controller.authentication_controller.login_with_user_code(pin_code=request_body['code'], accept_terms=True)
```
LnL7, line 872: I think it would be better to handle `accept_terms` implicitly (for now?) when registering users/rfid tags instead of during the authentication flow.

LnL7 (approved):

PR #712, opened 2021-06-09
Bugfix/v1 api inconcistencies

ptuytsch:


src/gateway/delivery_controller.py
```diff
@@ -114,27 +124,37 @@ def current_timestamp_to_string_format(cls):
...
         if delivery_dto.timestamp_pickup is not None:
             raise RuntimeError('Cannot update the delivery with id: {}: Delivery has already been picked up'.format(delivery_id))
 
         delivery_dto.timestamp_pickup = DeliveryController.current_timestamp_to_string_format()
-        return DeliveryController.save_delivery(delivery_dto)
+        if delivery_dto.type == Delivery.DeliveryType.RETURN:
+            pickup_user_dto = delivery_dto.user_pickup
+            delivery_dto.user_pickup = delivery_dto.user_delivery
+            delivery_dto_saved = self.save_delivery(delivery_dto)
+            self.user_controller.remove_user(pickup_user_dto)
```
LnL7, line 140: I don't really understand the logic here, do we create temporary users for each return?

src/gateway/user_controller.py
```diff
@@ -108,12 +115,18 @@ def load_user(self, user_id):
...
+    def load_users(self, roles=None, include_inactive=False):
+        # type: (List[str], bool) -> List[UserDTO]
         """  Returns a list of UserDTOs with all the usernames """
         _ = self
         users = []
-        for user_orm in User.select():
+        query = User.select()
+        if roles is not None:
+            clauses = [(User.role == role) for role in roles]
+            query = query.where(reduce(operator.or_, clauses))
```
LnL7, line 126: Isn't there a builtin query like `query.where(User.role.in(roles))`?

src/gateway/webservice_v1.py
```diff
@@ -344,14 +347,14 @@ def post_user(self, role, request_body):
 
         user_dto.username = uuid.uuid4().hex
         # add a custom user code
-        user_dto.pin_code = str(self._user_controller.generate_new_pin_code())
+        user_dto.pin_code = str(self._user_controller.generate_new_pin_code()).rjust(UserController.PinCodeLength[user_dto.role], '0')
```
LnL7, line 350: This seems to imply that the generation of new pin codes isn't long enough?

src/gateway/delivery_controller.py
ptuytsch: yes, we do. It is to generate a pickup code that is not already in use. When the delivery is picked up, the temporary user needs to be deleted. Would there be a better method to generate a pickup code that is not in the user.user_code column?

src/gateway/webservice_v1.py
ptuytsch: yes, that is the reason :) the generate_new_pin_code function will generate an integer, but it can be less than 4 or 6 digits, therefore the string is padded. I will change the generate function to take in this parameters, this will make more sense

src/gateway/user_controller.py
ptuytsch: it does! didn't know that, Thx!

LnL7 (approved):

PR #714, opened 2021-06-16
OM-2042 master bootloader timings

Shinigami-shinto:


src/gateway/hal/master_controller_classic.py
```diff
@@ -1208,41 +1208,46 @@ def update_master(self, hex_filename, controller_serial=INJECTED):
...
+                    # >
+                    # > Using /dev/ttyO5 at 115200 bps
+                    # > Connecting...
+                    # > Bootloader Firmware v1.05
+                    # > PIC18F67J11 Revision 10
+                    match = re.findall(pattern=r'Bootloader Firmware (v[0-9]+\.[0-9]+).*(PIC.*) Revision',
+                                       string=response,
+                                       flags=re.DOTALL)
+                    if not match:
+                        raise RuntimeError('Bootloader response did not match: {0}'.format(response))
```
khenderick, line 1244: Raising this exception means that if the first timing doesn't work, you'll won't try again.

I think you'll need some kind of `bootloader_active` variable (`False` by default) that you can set to `True` right before you break out of the loop. Instead of raising the exception, you log a warning, `continue` the loop and then below the loop you check the boolean and raise if not `True`.

khenderick (changes requested):

src/gateway/hal/master_controller_classic.py
Shinigami-shinto: Feedback aplied

khenderick (approved):

PR #715, opened 2021-06-23
Thermostat (and other) fixes

khenderick:


src/gateway/input_controller.py
```diff
@@ -123,7 +124,12 @@ def load_inputs(self):  # type: () -> List[InputDTO]
         inputs_dtos = []
         for input_ in list(Input.select(Input, Room)
                                 .join_from(Input, Room, join_type=JOIN.LEFT_OUTER)):  # TODO: Load dicts
-            input_dto = self._master_controller.load_input(input_id=input_.number)
+            try:
+                input_dto = self._master_controller.load_input(input_id=input_.number)
+            except TypeError as ex:
+                logger.error('Could not load input {0}: {1}'.format(input_.number, ex))
+                self.request_sync_orm()  # This likely is caused by module reconfigurations
```
LnL7, line 131: Not sure if this is a great idea, a but or misconfigured input could potentially cause constant syncing.

LnL7 (approved):

src/gateway/input_controller.py
khenderick: Valid point, I'll remove the statement. The scheduled sync will solve it a bit later anyway.

LnL7 (approved):

PR #716, opened 2021-06-23
Heartbeat TCP connection reuse

svanscho:


src/vpn_service.py
```diff
@@ -141,9 +144,9 @@ def call_home(self, extra_data):
         if self._url is None:
             logger.debug('Cloud not configured, skipping call home')
         try:
-            request = requests.post(self._url,
-                                    data={'extra_data': json.dumps(extra_data, sort_keys=True)},
-                                    timeout=10.0)
+            request = self._session.post(self._url,
+                                         data={'extra_data': json.dumps(extra_data, sort_keys=True)},
+                                         timeout=10.0)
```
LnL7, line 149: The request timeout can also be configured on the session, ensuring that all calls will timeout.

LnL7 (approved):
Did you check that the connections are reused between requests? It's not entirely clear to me how or how long urllib3 decides to keep connections open.

PR #718, opened 2021-06-28
OM-2289 Proceed with flashing the master firmware in case of a failed version call

Shinigami-shinto:


src/openmotics_update.py
```diff
@@ -290,17 +290,28 @@ def update_master_firmware(master_type, hexfile, version):
     elif master_type == 'master_coreplus':
         arguments += ['--master-firmware-core', hexfile]
     try:
+        # If the master is stuck in BL, the version call fails, but we should still be able to flash the master
         output = subprocess.check_output(['python', master_tool, '--version'])
+    except Exception as exc:
+        logger.exception('Checking master firmware version failed - trying to flash a couple of times')
+        for i in range(3):
```
svanscho, line 297: perhaps we can log the attempt number so it's clear how many times we will retry this in case of troubleshooting

khenderick (dismissed):

khenderick (approved):

PR #723, opened 2021-06-30
Adding support for shutter integration tests

khenderick:


testing/cicd/tests/hardware_layout.py
```diff
@@ -116,6 +133,10 @@ def __init__(self, name, mtype, hardware_type, is_can=False, inputs=None, cts=No
         for temp in (temps or []):
             temp.module = self
             self.temps.append(temp)
+        self.shutters = []
+        for shutter in (shutters or []):
```
Shinigami-shinto, line 137: I didn't understand this at first, but I guess it's to avoid: `NoneType is not iterable`

Shinigami-shinto (approved):
I went through it commit by commit in pycharm :)

testing/cicd/tests/hardware_layout.py
khenderick: Correct.

PR #729, opened 2021-07-05
OM-2316 Factory reset disables Large Installation

Shinigami-shinto:


khenderick (changes requested):
Already fixed at https://github.com/openmotics/gateway/blob/develop/src/gateway/hal/master_controller_classic.py#L1356

This was also an issue when doing a restore.

PR #732, opened 2021-07-12
Feature/esafe events

ptuytsch:


src/gateway/pubsub.py
```diff
@@ -93,6 +103,14 @@ def _publish_all_events(self):
                 self._publish_gateway_event(*event)
             except Empty:
                 break
+        while True:
+            try:
+                event = self._esafe_events.get(block=True, timeout=0.25)
+                if event is None:
+                    return
```
khenderick, line 110: I think I just noticed a bug; not on this line, but on the two return statements above. If there's no e.g. master event after two seconds, it should rather break the while loop and try the next event types. /CC @LnL7

khenderick (changes requested):

src/gateway/pubsub.py
ptuytsch: isn't this the case with the break statement when an empty exception is raised due to the events queue being empty (with the timeout of 0.25 seconds)? I'll test this in more detail to be sure.

khenderick: Nevermind, my bad. The `stop` function puts `None` into these queues to abort the thread.

khenderick (approved):

PR #733, opened 2021-07-13
Reflecting event changes

khenderick:


src/master/core/events.py
```diff
@@ -149,39 +183,102 @@ def data(self):
...
+                                   'value': Temperature.system_value_to_temperature(self._data[1])}]
             elif self._action == 1:
-                sensor_type = Event.SensorType.HUMIDITY
-                sensor_value = Humidity.system_value_to_humidity(self._data[1])
+                sensor_values += [{'type': Event.SensorType.HUMIDITY,
+                                   'value': Humidity.system_value_to_humidity(self._data[1])}]
             elif self._action == 2:
-                sensor_type = Event.SensorType.BRIGHTNESS
-                sensor_value = self._word_decode(self._data[0:2])
+                sensor_values += [{'tye': Event.SensorType.BRIGHTNESS,
```
LnL7, line 212:
```suggestion
                sensor_values += [{'type': Event.SensorType.BRIGHTNESS,
```

src/master/core/events.py
```diff
@@ -149,39 +183,102 @@ def data(self):
...
+                                   'value': self._word_decode(self._data[0:2])}]
+            elif self._action == 3:
+                sensor_values += [{'type': Event.SensorType.CO2,
```
LnL7, line 215: FYI while events for new sensor types are handled here the sensor controller still won't sync the new types. Some changes are still needed there to fully support co2/voc sensors.

src/master/core/events.py
khenderick: Nice catch

src/master/core/events.py
khenderick: Good point, I've made sure the SensorController wouldn't raise exceptions if these new sensor types would be send.

LnL7 (approved):

PR #739, opened 2021-07-19
Fixed leak in plugin runners

khenderick:
Symptoms:
* `ps faux | grep python`: A lot of identical plugins running
* `ls -al /proc/<pid openmotics_service.py>/fd/*`: A lot of open file descriptors, most of them pipes towards the runners

### DIFF
--- a/src/plugin_runtime/runtime.py
+++ b/src/plugin_runtime/runtime.py
@@ -400,6 +400,7 @@ class PluginRuntime(object):
         except Exception as exception:
             return {'success': False, 'exception': str(exception), 'stacktrace': traceback.format_exc()}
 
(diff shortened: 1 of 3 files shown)


svanscho (dismissed):

svanscho:
Approved. Not sure how much work it would be to add a test to capture the condition to always stop the plugin.

khenderick:
> Approved. Not sure how much work it would be to add a test to capture the condition to always stop the plugin in case there are exceptions.
Added a unit test to cover this. However, might still be a good idea to add a few plugin related integration tests as well.

LnL7 (approved):

PR #744, opened 2021-07-23
Make dynamic updates more flexible

khenderick:


testing/cicd/tests/conftest.py
```diff
@@ -78,53 +78,66 @@ def firmware_updates(toolbox_session):
...
+                    firmware[firmware_type] = desired_version
+            else:
+                logger.info('Skipping {0} - no module found'.format(firmware_type))
+
+    if TEST_PLATFORM == TestPlatform.DEBIAN:
+        _check_and_schedule('M', 'master_classic', os.environ.get('OPENMOTICS_MASTER_CLASSIC_FIRMWARE'))
+    if TEST_PLATFORM == TestPlatform.CORE_PLUS:
+        _check_and_schedule('M', 'master_core', os.environ.get('OPENMOTICS_MASTER_CORE_FIRMWARE'))
+
+    _check_and_schedule('O', 'output', os.environ.get('OPENMOTICS_OUTPUT_FIRMWARE'))
```
LnL7, line 103: Do we want to use the same variables here? Explicitly differentiating between gen2/gen3 could avoid some unexpected behaviour when invalid parameters are used.

testing/cicd/tests/toolbox.py
```diff
@@ -395,7 +395,7 @@ def get_gateway_version(self):
 
     def get_firmware_versions(self):
         # type: () -> Dict[str,str]
-        modules = self.dut.get('/get_modules_information')['modules']['master']
+        modules = self.dut.get('/get_modules_information?refresh=True')['modules']['master']
```
LnL7, line 398: This should already be refreshing explicitly when firmware is flashed.

khenderick: The call returns data from the ORM. And while we typically refresh the data in the ORM after certain events (startup, updates, ...) it would still be possible for the ORM to be outdated at the moment the update test starts and wants to fetch the pre-test state (to decide what the tests can and should update). So it's not that it would be used in the wild, but for the tests, I think it adds some "stability"

testing/cicd/tests/conftest.py
khenderick: As discussed, the update logic covers the case where one would try to update a gen3 module with gen2 firmware and vice versa. And since for now a single automated test job can only handle one system at a time, sharing these is OK I think.

LnL7 (dismissed):

LnL7 (approved):

PR #751, opened 2021-07-28
Feature/impersonate login

ptuytsch:
Things changed:
 - Make use of the SUPER user role
 - Add the impersonate argument to the API to be able to impersonate as a different user.
   - This only works as a SUPER user, no other user roles can use this feature
 - Make the authentication token store more data:
   - Login method: Password, pin_code or RFID
   - impersonator: When the token is from a impersonated ("SUPER") account, the ("SUPER") user that requested the login is stored in the token for future reference.
 - Return the username field of a user in the users API. This can be used for the impersonate field when logging in.
 - Separate the User-controller and the authentication-controller for cleaner code. (was more strongly coupled before)


src/gateway/authentication_controller.py
```diff
@@ -72,26 +89,49 @@ def login(self, user_dto, accept_terms=False, timeout=None):
...
+        # Check if accepted terms
         if user_orm.accepted_terms == AuthenticationController.TERMS_VERSION:
-            token = self.token_store.create_token(UserMapper.orm_to_dto(user_orm), timeout=timeout)
+            token = self.token_store.create_token(user_to_login, timeout=timeout, impersonator=impersonator, login_method=login_method)
             return True, token
         if accept_terms is True:
             user_orm.accepted_terms = AuthenticationController.TERMS_VERSION
             user_orm.save()
-            token = self.token_store.create_token(UserMapper.orm_to_dto(user_orm), timeout=timeout)
+            token = self.token_store.create_token(user_to_login, timeout=timeout, login_method=login_method)
```
yclybouw, line 129: Shouldn't the `impersonator` be added to? Because the `user_to_login` might be impersonated.

src/gateway/authentication_controller.py
```diff
@@ -72,26 +89,49 @@ def login(self, user_dto, accept_terms=False, timeout=None):
...
             return True, token
         return False, UserEnums.AuthenticationErrors.TERMS_NOT_ACCEPTED
 
-    def login_with_user_code(self, pin_code, accept_terms=False, timeout=None):
-        # type: (str, bool, Optional[float]) -> Tuple[bool, Union[str, AuthenticationToken]]
+    def login_with_user_code(self, pin_code, accept_terms=False, timeout=None, impersonate=None):
```
yclybouw, line 133: I would not allow this with pin_code.

src/gateway/authentication_controller.py
```diff
@@ -101,20 +141,21 @@ def login_with_user_code(self, pin_code, accept_terms=False, timeout=None):
             return False, UserEnums.AuthenticationErrors.INVALID_CREDENTIALS
 
         user_dto = UserMapper.orm_to_dto(user_orm)
-        return self.login(user_dto, accept_terms=accept_terms, timeout=timeout)
+        return self.login(user_dto, accept_terms=accept_terms, timeout=timeout, login_method=LoginMethod.PIN_CODE, impersonate=impersonate)
 
-    def login_with_rfid_tag(self, rfid_tag_string, accept_terms=False, timeout=None):
-        # type: (str, bool, Optional[float]) -> Tuple[bool, Union[str, AuthenticationToken]]
+    def login_with_rfid_tag(self, rfid_tag_string, accept_terms=False, timeout=None, impersonate=None):
```
yclybouw, line 146: I would not allow this wit RFID (only trough the web, thus through username/password).

src/gateway/authentication_controller.py
```diff
@@ -164,23 +205,26 @@ def remove_token_for_user(self, user_dto):
...
 
-    def create_token(self, user_dto, timeout=None):
-        # type: (UserDTO, int) -> AuthenticationToken
+    def create_token(self, user_dto, login_method, timeout=None, impersonator=None):
+        # type: (UserDTO, LoginMethod, int, Optional[UserDTO]) -> AuthenticationToken
         """ Creates an authentication token, and needs the user code to do so """
         full_user = self._get_full_user_dto(user_dto)
         user_id = full_user.id
         # if there is already a token, just return the existing token
         if user_id in self.tokens:
```
yclybouw, line 217: This cannot work any more. E.g. a user could be logged in on the display through PIN code, and in the meantime a superuser impersonates for that user. The tokens should be different (e.g. for logging, or for access rights).

src/gateway/authentication_controller.py
```diff
@@ -164,23 +205,26 @@ def remove_token_for_user(self, user_dto):
...
             return self.tokens[user_id]
         if timeout is not None:
-            self.tokens[user_id] = AuthenticationToken.generate(full_user, token_timeout=timeout)
+            self.tokens[user_id] = AuthenticationToken.generate(full_user, token_timeout=timeout, impersonator=impersonator, login_method=login_method)
         else:
-            self.tokens[user_id] = AuthenticationToken.generate(full_user)
+            self.tokens[user_id] = AuthenticationToken.generate(full_user, impersonator=impersonator, login_method=login_method)
+        logger.info(' -*****- New token created: {}'.format(self.tokens[user_id].to_dict()))
```
yclybouw, line 223: Please remove before merging :-)

src/gateway/webservice_v1.py
```diff
@@ -899,14 +899,20 @@ def __init__(self):
     def authenticate_pin_code(self, request_body):
         if 'code' not in request_body:
             raise WrongInputParametersException('Expected a code in the request body json')
-        success, data = self._user_controller.authentication_controller.login_with_user_code(pin_code=request_body['code'])
+        impersonate_username = None
+        if 'impersonate' in request_body:
```
yclybouw, line 903: I would not allow impersonation for PIN-login.

src/gateway/webservice_v1.py
```diff
@@ -899,14 +899,20 @@ def __init__(self):
...
+        success, data = self.authentication_controller.login_with_user_code(pin_code=request_body['code'], impersonate=impersonate_username)
         return self.handle_authentication_result(success, data)
 
     @openmotics_api_v1(auth=False, expect_body_type='JSON')
     def authenticate_rfid_tag(self, request_body):
         if 'rfid_tag' not in request_body:
             raise WrongInputParametersException('Expected an rfid_tag in the request body json')
-        success, data = self._user_controller.authentication_controller.login_with_rfid_tag(rfid_tag_string=request_body['rfid_tag'])
+        impersonate_username = None
+        if 'impersonate' in request_body:
```
yclybouw, line 913: I would not allow impersonation for RFID-login.

testing/unittests/api_tests/V1/rfid_test.py
```diff
@@ -91,39 +93,39 @@ def setUp(self):
 
     def test_get_rfid(self):
         with mock.patch.object(self.rfid_controller, 'load_rfids', return_value=self.all_test_rfids):
-            auth_token = AuthenticationToken(self.test_admin_1, token='test-token', expire_timestamp=int(time.time() + 3600))
+            auth_token = AuthenticationToken(self.test_admin_1, token='test-token', expire_timestamp=int(time.time() + 3600), login_method=LoginMethod.PASSWORD)
```
yclybouw, line 96: Shouldn't the login method be RFID in this file?

yclybouw (changes requested):

src/gateway/authentication_controller.py
ptuytsch: Good catch, I changed the key of the dict to an custom object that keeps track of the logged in user and the impersonator to overcome this problem

yclybouw (approved):

PR #754, opened 2021-07-30
OM-2318 sensors fix

svanscho:
Return sensor object in set_sensor_configuration(s) API calls

### DIFF
--- a/testing/unittests/gateway_tests/webservice_test.py
+++ b/testing/unittests/gateway_tests/webservice_test.py
@@ -20,6 +20,7 @@ import unittest
 import mock
 
 from bus.om_bus_client import MessageClient
+from gateway.api.serializers import SensorSerializer
 from gateway.dto import DimmerConfigurationDTO, LegacyScheduleDTO, \
     LegacyStartupActionDTO, ModuleDTO, OutputStatusDTO, ScheduleDTO, \
     SensorDTO, SensorSourceDTO, SensorStatusDTO, UserDTO, VentilationDTO, \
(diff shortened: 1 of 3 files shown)


testing/unittests/gateway_tests/webservice_test.py
```diff
@@ -175,15 +176,29 @@ def test_sensor_configurations(self):
                 'virtual': False,
             }], json.loads(response)['config'])
 
+    def test_set_sensor_configuration(self):
```
LnL7, line 179: Is there also a test for the registration case? (post without id and response with)

LnL7 (approved):

testing/unittests/gateway_tests/webservice_test.py
svanscho: There is a test in the sensor controller:
https://github.com/openmotics/gateway/blob/60f480b2d12a08dce7451e16be97225797a5f707/testing/unittests/gateway_tests/sensor_controller_test.py#L248
From an E2E perspective, would rather not test in the webservice as it requires some refactoring with not mocking the sensor controller any longer etc.

LnL7: these are not webservice tests but controller tests

LnL7: checking the id of the return dto in the return value here would cover it https://github.com/openmotics/gateway/blob/60f480b2d12a08dce7451e16be97225797a5f707/testing/unittests/gateway_tests/sensor_controller_test.py#L255.

PR #764, opened 2021-08-04
Feature/detailed deliveries api

ptuytsch:
Add new functionallity to the delivery api:
 - Add the delivery history call to get a paginated result of the delivery history
 - Add a query parameter to the parcelbox to nest the delivery result when applicable for the parcelbox.
 - some minor improvements in the controllers

### DIFF
--- a/src/gateway/api/V1/parcelbox.py
+++ b/src/gateway/api/V1/parcelbox.py
@@ -30,6 +30,8 @@ from gateway.models import User
 from gateway.webservice_v1 import RestAPIEndpoint, openmotics_api_v1, expose
 
 if False:  # MyPy
+    from typing import Optional
+    from gateway.authentication_controller import AuthenticationToken
     from gateway.delivery_controller import DeliveryController
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


src/gateway/delivery_controller.py
```diff
@@ -84,9 +75,37 @@ def load_deliveries(user_id=None, include_picked_up=False):
                  (Delivery.user_pickup_id == user_id))
             )
         # filter on picked up when needed
+        query = query.where(Delivery.timestamp_pickup.is_null(not history))
+
+        # add the from_id
+        query = query.where(Delivery.id > from_id)
```
yclybouw, line 81: As the history is shown in descending order (which makes sense), the "after" or "from_id" should be the upper limit, so: `query.where(Delivery.id < from_id)`

yclybouw (changes requested):

src/gateway/delivery_controller.py
yclybouw: This reminds me: from_id should not default to `0` but to "infinity". E.g. set to `None` and do the where conditionally.

ptuytsch: indeed good catch! The fix has been pushed

yclybouw (approved):

yclybouw (approved):

src/gateway/api/V1/parcelbox.py
```diff
@@ -63,13 +63,17 @@ def __init__(self, esafe_controller=INJECTED, delivery_controller=INJECTED):
                                       controller=self, action='put_parcelbox',
                                       conditions={'method': ['PUT']})
 
-    @openmotics_api_v1(auth=False, expect_body_type=None, check={'size': str, 'available': bool})
-    def get_parcelboxes(self, size=None, available=None):
+    @openmotics_api_v1(auth=False, expect_body_type=None, check={'size': str, 'available': bool, 'show_deliveries': bool})
+    def get_parcelboxes(self, size=None, available=None, show_deliveries=None):
```
LnL7, line 67: Any user can list pending delivery information?

LnL7 (dismissed):

src/gateway/api/V1/parcelbox.py
ptuytsch: Good point, will fix this.

LnL7 (approved):

PR #768, opened 2021-08-06
Bugfix/user iso language

ptuytsch:


src/gateway/enums.py
```diff
@@ -137,3 +149,12 @@ class SerialPorts(object):
     ENERGY = 'ENERGY'
     P1 = 'P1'
     EXPANSION = 'EXPANSION'
+
+
+class Languages(BaseEnum):
+    """ languages, ISO 639-1 format"""
+    # only the current supported languages are included
+    EN = 'en'
+    DE = 'DE'
```
khenderick, line 158: I think the `DE` must be lowercase here.

src/gateway/migrations/orm/024_convert_language_to_iso.py
```diff
@@ -0,0 +1,96 @@
...
+        'Francais': 'fr'
+    }
+    for user in User.select():
+        try:
+            old_lang = user.language
+            user.language = language_translation[old_lang]
+            user.save()
+        except KeyError:
+            pass
+
```
khenderick, line 92: I miss a migration to update the default value for the `language` field.

src/gateway/migrations/orm/024_convert_language_to_iso.py
```diff
@@ -0,0 +1,96 @@
...
+    language_translation = {
+        'English': 'en',
+        'Deutsch': 'de',
+        'Nederlands': 'nl',
+        'Francais': 'fr'
+    }
+    for user in User.select():
+        try:
+            old_lang = user.language
+            user.language = language_translation[old_lang]
```
khenderick, line 88: I think you should better use `language_translation.get(old_lang, 'en')` to make sure the migration will actually work in case anyone would have manually changed the language in the DB and made a typo

khenderick (changes requested):

khenderick (approved):

PR #775, opened 2021-08-12
Metrics format regression after energy refactor

svanscho:


src/gateway/metrics_collector.py
```diff
@@ -671,7 +671,7 @@ def _add_if_not_none(dictionary, field, value):
         energy_data = {}
         try:
             for energy_module in self._energy_module_controller.load_modules():
-                device_id = '{0}.{{0}}'.format(energy_module.address)
+                device_id = 'E{0}.{{0}}'.format(energy_module.address)
```
khenderick, line 674: Nice catch. However, you should just use `energy_module.formatted_address` instead, which will also handle the P1C case.

khenderick (changes requested):

khenderick:
I had to fix the tests, as it seems the tests previously used a slightly different/reduced code path in the tests, where the formatting didn't happen. Now that the tests follow a more accurate code path, the `device_id` is also updated.

khenderick (approved):

khenderick:
Self-approved after review by @svanscho

PR #776, opened 2021-08-12
Bugfix/frontend linking mismatches

ptuytsch:


src/gateway/migrations/orm/025_apartment_rebus_id_nullable.py
```diff
@@ -0,0 +1,59 @@
...
+        doorbell_rebus_id = IntegerField(unique=True, null=True)
+
+    # Set both rebus id fields to be nullable
+    # This does not work for some reason, instead deleting the table and creating the table again
+    # to include the null=True change
+    # migrator.change_columns(Apartment, mailbox_rebus_id=IntegerField(null=True, unique=True, default=None))
+    # migrator.change_columns(Apartment, doorbell_rebus_id=IntegerField(null=True, unique=True, default=None))
+
+    # Can be done, no apartments in use at this moment
+    migrator.drop_table(Apartment)
```
khenderick, line 53: It's ok if we're sure that we don't have any valuable data in this table at the moment of release. Otherwise, we could probably have a double migration, where we some field, move the data and remove the old one, and then have another that adds the correct field, moves the data and remove the temporarily field again.

khenderick (approved):

PR #784, opened 2021-08-24
Bugfix/cherrypy websockets

ptuytsch:


src/gateway/webservice.py
```diff
@@ -236,7 +237,7 @@ def authentication_handler(pass_token=False, pass_role=False, version=0):
             else _self.webinterface.check_token
         checked_token = check_token(token)  # type: Optional[AuthenticationToken]
         # check if the call is done from localhost, and then verify the token
-        if request.remote.ip != '127.0.0.1':
+        if Config.get_entry('http_enable', True) and request.remote.ip != '127.0.0.1':
```
khenderick, line 240: So if http is disabled (and only https is enabled), authentication is turned off? Not sure whether that's on purpose.

khenderick (changes requested):

khenderick (approved):

PR #795, opened 2021-09-03
Fix cold reset on the Brain

khenderick:


src/platform_utils.py
```diff
@@ -124,21 +128,37 @@ def get_mac_address():  # type: () -> Union[str, None]
...
     def set_gpio(gpio_pin, value):  # type: (int, bool) -> None
         Hardware.set_gpio_direction(gpio_pin=gpio_pin,
                                     direction=Hardware.GPIO_DIRECTION.OUT)
+        if gpio_pin in Hardware.INVERTED_GPIOS:
+            value = not value
         with open(Hardware.GPIO_VALUE_PATH.format(gpio_pin), 'w') as gpio:
             gpio.write('1' if value else '0')
 
+    @staticmethod
+    def cycle_gpio(gpio_pin, cycle):  # type: (int, List[Union[bool, float]]) -> None
```
LnL7, line 147: This could a use a description since the arguments/implementation isn't immediately obvious.

LnL7 (dismissed):

src/platform_utils.py
khenderick: Valid remark, added some docs with an example.

LnL7 (approved):

PR #797, opened 2021-09-03
Feature/long can bus

Shinigami-shinto:


testing/cicd/tests/toolbox.py
```diff
@@ -457,7 +456,10 @@ def discover_modules(self, output_modules=False, input_modules=False, shutter_mo
                 self.tester.toggle_output(TESTER.Button.can, delay=0.5)
                 module_amounts = {'C': 1}
                 if ucans:
-                    module_amounts.update({'I': 1, 'T': 1})
+                    # We will have 3 emulated Input modules and 2 emulated Temp modules
```
khenderick, line 459: Not sure about the comment. In theory, you should press the CC button once after all uCANs would be "discovered" by it. So basically; press at least one input at every uCAN, and afterwards, press the CC button once, after which it will with a small interval announce all it's emulated modules.

khenderick, line 459: Not sure about the comment. In theory, you should press the CC button once after all uCANs would be "discovered" by it. So basically; press at least one input at every uCAN, and afterwards, press the CC button once, after which it will with a small interval announce all it's emulated modules.

khenderick:
Will be merged via #813

PR #799, opened 2021-09-09
Bugfix/cloud link bugfixes

ptuytsch:
Various small bugfixes discovered during linking cloud to the eSafe version of the gateway.
 - Adding more dummy eSafe parcelboxes for testing purposes
 - Show the pin code for the user when requested with the correct authentication
 - return a 401 status when an invalid auth token is passed, instead of handling the call without any authentication
 - make the is_active field editable with the user put call
 - when requesting the parcel history, also include the non picked up packages
 - add the v1_api to the features list

### DIFF
--- a/src/gateway/webservice.py
+++ b/src/gateway/webservice.py
@@ -670,6 +670,7 @@ class WebInterface(object):
             'websocket_maintenance',  # Maintenance over websockets
             'shutter_positions',  # Shutter positions
             'ventilation',
+            'v1_api'
         ]
 
         master_version = self._module_controller.get_master_version()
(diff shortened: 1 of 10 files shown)


src/gateway/webservice.py
```diff
@@ -670,6 +670,7 @@ def get_features(self):
             'websocket_maintenance',  # Maintenance over websockets
             'shutter_positions',  # Shutter positions
             'ventilation',
+            'v1_api'
```
LnL7, line 673: Since this is very generic, I wonder if it might be better to `'esafe'` or something similar instead. This also could be added conditionally based on actual esafe feature availability so it's not included on openmotics hardware.

LnL7 (approved):

PR #801, opened 2021-09-10
thermostats eeprom migration

LnL7:


src/gateway/thermostat/master/thermostat_controller_master.py
```diff
@@ -48,17 +50,19 @@ class ThermostatControllerMaster(ThermostatController):
...
+        # type: (OutputController, MasterClassicController, EepromController, PubSub) -> None
         super(ThermostatControllerMaster, self).__init__(output_controller)
         self._master_controller = master_controller  # classic only
+        self._eeprom_controller = eeprom_controller  # classic only
         self._pubsub = pubsub
 
         self._monitor_thread = DaemonThread(name='thermostatctl',
                                             target=self._monitor,
-                                            interval=1, delay=10)
+                                            interval=30, delay=10)
```
khenderick, line 65: This will add a significant delay to thermostats events, what's the usecase of increasing this? I would expect that if master thermostats are disabled, this controller wouldn't be loaded/started altogether.

Since the migration of thermostats takes place at startup, I would expect that the `start` method would just check the eeprom and refuse to start if it was disabled.

src/openmotics_service.py
```diff
@@ -185,6 +188,12 @@ def start(master_controller=INJECTED,  # type: MasterController
         EnergyModulesMigrator.migrate()
         EsafeMigrator.migrate()
 
+        thermostats_gateway_enabled = Feature.select(Feature.enabled) \
```
khenderick, line 191: Maybe add a comment indicating why the filter is here and not (like for example the platform) inside the migrate method. To make sure that nobody moves it inside and then mark the migration as completed on all gateways when the feature isn't enabled yet.

khenderick (changes requested):

src/gateway/thermostat/master/thermostat_controller_master.py
LnL7: The actual master thermostat logic only runs every 1 or 2 minutes so polling every second seems very overkill to me.

khenderick: Unless you make an actual change (e.g. change mode), then it updates everything (e.g. new setpoints according to this mode) immediately as far as I known. This will now take up to 30 seconds.

khenderick (approved):

PR #812, opened 2021-09-23
Remove sensor cache expiry

khenderick:


src/gateway/dto/sensor.py
```diff
@@ -52,11 +52,11 @@ def is_plugin(self):
...
 
 class SensorStatusDTO(BaseDTO):
-    def __init__(self, id, value=None, last_value=None):
+    def __init__(self, id, value=None, last_updated=None):
         # type: (int, Optional[float], Optional[float]) -> None
         self.id = id
-        self.value = float(value) if value else None
-        self.last_value = last_value
+        self.value = float(value) if value is not None else None
+        self.value_last_updated = last_updated
```
LnL7, line 59: This shouldn't really be needed anymore.

LnL7 (approved):

src/gateway/dto/sensor.py
khenderick: Hmm, right, I'll remove it right away

LnL7 (approved):

PR #816, opened 2021-09-28
Small thermostat_group cleanup

khenderick:


src/gateway/thermostat/gateway/thermostat_controller_gateway.py
```diff
@@ -293,11 +296,16 @@ def get_temperature_from_sensor(sensor):  # type: (Optional[Sensor]) -> Optional
             statuses.append(group_status)
         return statuses
 
-    def set_thermostat_group(self, thermostat_group_id, group_on, cooling_mode=False, cooling_on=False, automatic=None, setpoint=None):
-        # type: (int, bool, bool, bool, Optional[bool], Optional[int]) -> None
-        mode = ThermostatMode.COOLING if cooling_mode else ThermostatMode.HEATING  # type: Literal['cooling', 'heating']
+    def set_thermostat_mode(self, thermostat_on, cooling_mode=False, cooling_on=False, automatic=None, setpoint=None):
+        self.set_thermostat_group(thermostat_group_id=0,
+                                  state=ThermostatState.ON if thermostat_on or cooling_on else ThermostatState.OFF,
+                                  mode=ThermostatMode.COOLING if cooling_mode or cooling_on else ThermostatMode.HEATING)
```
LnL7, line 302: I'm not sure this logic is correct, some api interactions might assume that `cooling_mode=False cooling_on=True` is a valid way to switch to heating mode.

LnL7 (approved):

src/gateway/thermostat/gateway/thermostat_controller_gateway.py
khenderick: Valid point, new implementation should be more correct and in line with what one could expect.

LnL7 (approved):

PR #848, opened 2021-10-21
attempt to restore the previously scheduled auto setpoint

LnL7:
When switching to the auto preset any overridden setpoints will be
restored back to the setpoint that would have been scheduled for the
current time window.


LnL7:
Replaced by https://github.com/openmotics/gateway/pull/927.

PR #849, opened 2021-10-22
Bugfix/esafe user apartment

yclybouw:
This will change the API so that an apartment id is not a list of apartments but one single integer that represents the id of the apartment that is linked to that user.
This goes together with the required changes in the front-end in pull request: https://github.com/openmotics/esafe-front-end/pull/29
furthermore, the serializer tests have been relocated to the api/serializer folder


ptuytsch:
Code has been seen by Yannick and the changes have been approved. Yannick cannot approve since he is the original author of the PR

ptuytsch (approved):

PR #857, opened 2021-11-02
CC sync

khenderick:


src/gateway/hal/master_controller_core.py
```diff
@@ -1151,16 +1151,12 @@ def _get_version(address_):
...
             can_control_module_info = CanControlModuleConfiguration(module_id)
             device_type = can_control_module_info.device_type
-            hardware_type = HardwareType.PHYSICAL
-            if module_id == 0:
-                hardware_type = HardwareType.INTERNAL
             dto = ModuleDTO(source=ModuleDTO.Source.MASTER,
                             address=can_control_module_info.address,
                             module_type=module_type_lookup.get(device_type.lower()),
-                            hardware_type=hardware_type,
+                            hardware_type=HardwareType.PHYSICAL,
```
Shinigami-shinto, line 1157: Can you give some context about hardware_types and gen3?
I'm guessing only the internal input and output modules of the brain+ have the HardwareType.INTERNAL type? The CC is now considered Physical, what's the reason behind this? :)

Shinigami-shinto (commented):
I have a question about the first commit, the 2nd commit makes sense as the update controller now handles that logic

src/gateway/hal/master_controller_core.py
khenderick: Previously the build-in CC module from the Brain(+) would also be added to the ORM. However, that's not the case anymore, since we consider it the Brain(+) itself, and not some kind of "internal CC module" anymore, it's the Brain(+) itself. This means that the only CC that exists are physical ones that are just attached as usual to the RS485 bus.

Shinigami-shinto (approved):

PR #900, opened 2021-11-29
Bugfix/slow rfid

ptuytsch:


src/gateway/pubsub.py
```diff
@@ -75,51 +104,44 @@ def start(self):
...
+        """ This function will continuously loop over the function to publish all the events in the queue"""
         while self._is_running:
             self._publish_all_events()
 
     def _publish_all_events(self, blocking=True):
         while True:
             try:
-                event = self._master_events.get(block=blocking, timeout=0.25)
-                if event is None:
+                event_structure = self._event_queue.get(block=blocking, timeout=0.25)
```
ptuytsch, line 118: here, it is cleaner to not set a timeout, but this if for testing purposes. It would be cleaner if there is no timeout and if the call is blocking for when an event got pushed into the queue.

testing/unittests/gateway_tests/pubsub_test.py
```diff
@@ -98,3 +98,58 @@ def test_pubsub_esafe_event_delivery(self):
...
+        # Test the time constraint
+        self.pubsub.publish_master_event(PubSub.MasterTopics.INPUT, master_event)
+        self.pubsub.publish_gateway_event(PubSub.GatewayTopics.STATE, gw_event)
+        self.pubsub.publish_esafe_event(PubSub.EsafeTopics.LOCK, esafe_event)
+
+        t_start = time.time()
+        self.pubsub._publish_all_events(blocking=True)  # Set the wait for end of queue to be true
+        t_end = time.time()
+        t_delta = t_end - t_start
+        self.assertLess(t_delta, 0.5)  # assert that all the events are send in less than half a second
```
ptuytsch, line 150: The previous mentioned timeout that is used in the pubsub code, is used here to test that all the events now get published in under half a second (this is 0.25 s margin of time). This will verify that the pubsub `_publish_all_events` function will take less than half a second, compared to previously 0.75 seconds.

ptuytsch: Is there a cleaner way to test this behavior or verify that the events do not take 0.75 seconds to pass the pubsub module?

src/gateway/pubsub.py
khenderick: Not only for testing purposes, but for service shutdown purposes; if the call would be permanently blocking, there was no way to stop this thread. Here it will take at most 0.25s to check for the self._is_running above again.

khenderick (approved):

PR #908, opened 2021-12-03
Added support for pulse counters on the Brain(+)

khenderick:


khenderick:
Awaiting `firmware-coreplus` version `v1.0.116` fixing an incorrect endianness.

src/gateway/hal/master_controller_core.py
```diff
@@ -869,21 +869,55 @@ def _legacy_brightness_to_lux(brightness):
...
+            if input_module_info.device_type == 'i' and '.000.000.' not in input_module_info.address:
+                continue  # Skip virtual modules
+            ids.append(module_id)
+        return ids
+
+    @staticmethod
+    def _generate_pulse_counter_dto(pulse_counter_id):
+        return PulseCounterDTO(id=pulse_counter_id,
+                               name='PulseCounter {0}'.format(pulse_counter_id),
+                               input_id=None,
```
svanscho, line 887: I would have expected the input_id to provided (although not configurable)

svanscho (changes requested):

src/gateway/hal/master_controller_core.py
khenderick: That was my initial implementation, but that would have more impact on user interfaces, since technically this means the input would be marked at "linked to a pulse counter" on multiple places, making the code on that part more complex. Setting it to `None` solves that.

svanscho (approved):

PR #912, opened 2021-12-03
BE-340 Support for energy related sensors

svanscho:


src/gateway/api/serializers/sensor.py
```diff
@@ -50,6 +56,11 @@ class SensorSerializer(object):
         Sensor.Units.LUX,
         Sensor.Units.MICRO_GRAM_PER_CUBIC_METER,
         Sensor.Units.PARTS_PER_MILLION,
+        Sensor.Units.VOLT,
+        Sensor.Units.AMPERE,
+        Sensor.Units.HERTZ,
+        Sensor.Units.KWH,
```
LnL7, line 62: Using an abbreviation is inconsistent with all the other units. Also have we decided what we want to do with modifiers like kilowatt and microgram? eg. all energy measurements on the gateway consistently use W/Wh.

svanscho: Agreed, then looked it up on wikipedia. The only SI supported energy measurement is Joule. This is not practical as most of our customers use energy measurements in kWh both for heat and electrical energy. In the NON-SI table, kWh is listed as unit, not Wh. As such I opted for kWh.

svanscho: The hour is a unit of time listed among the [non-SI units accepted](https://en.wikipedia.org/wiki/Non-SI_units_mentioned_in_the_SI) by the [International Bureau of Weights and Measures](https://en.wikipedia.org/wiki/International_Bureau_of_Weights_and_Measures) for use with the SI.[[3]](https://en.wikipedia.org/wiki/Kilowatt-hour#cite_note-BIPM-3) Its combination with the kilowatt, a standard SI unit, is therefore permitted within the standard.

LnL7 (approved):

PR #919, opened 2021-12-07
Send local time to cloud

khenderick:


svanscho:
Note sure this is the most ideal place. Why not sending it as system metrics? This is a more generic concept that can be reused throughout?

khenderick:
The benefit of sending it along with the rest of the metrics, is that you can very accurately calculate the internal delay (min/max difference between this timestamp and the timestamps of the metrics) of the gateway. If you'd for example pass the local time with the heartbeat, you'd have to make a general calculation of the time difference between gateway and cloud, and then between the metrics and the cloud.

PR #922, opened 2021-12-08
Added API to retry updates on inconsistent modules

svanscho:


src/gateway/webservice.py
```diff
@@ -2238,6 +2238,18 @@ def update(self, version, metadata=None):
         self._update_controller.request_update(version, metadata)
         return {}
 
+    @openmotics_api(auth=True, plugin_exposed=False)
+    def sync_inconsistent_modules(self, module_type=None):
```
khenderick, line 2242: Not sure if `sync_inconsistent_modules` covers what it does; what is "sync", and what are "inconsistent" modules? Maybe something like `retry_failed_firmware_updates` is more clear what will actually happen?

src/gateway/webservice.py
```diff
@@ -2238,6 +2238,18 @@ def update(self, version, metadata=None):
...
+        """ Request a firmware update on all modules that are not on the target version"""
+        for (module, target_version) in UpdateController.get_inconsistent_modules(module_type=module_type):
+            logger.info('Updating inconsistent module %s (%s) to %s',
+                        module.address,
+                        module.module_type,
+                        target_version)
+            module.update_success = None  # Allow the update to be re-tried
```
khenderick, line 2249: I'd rather have this loop and the update of `module.update_success` in the `UpdateController` so all logic about what this `update_success` actually is or does is in there. If we for example ever want to implement a single retry after a while or somehting, we might have to alter this field, and it would be nice if we only have to be in the `UpdateController`.

You can just do something like

```
self._update_controller.<whatever this function is called>(...)
return {}
```

src/gateway/update_controller.py
```diff
@@ -253,6 +253,28 @@ def request_update(self, new_version, metadata=None):
             global_logger.info('Request for update firmware {0} to {1}'.format(firmware_type, version))
         Config.set_entry('firmware_target_versions', target_versions)
 
+    @staticmethod
+    def get_inconsistent_modules(module_type=None):
```
khenderick, line 257: It's missing the retry of the master itself, which was the actual reason to implement these calls in the first place.

khenderick (changes requested):

khenderick:
If you would continue on these changes, make sure to incorporate the changes merged into develop via #923, especially setting the `self._pending_updates` to `True` if any of the update markers is (re)set.

PR #923, opened 2021-12-09
Increase robustness in Gen3 updates

khenderick:


src/gateway/update_controller.py
```diff
@@ -513,14 +519,19 @@ def _execute_pending_updates(self):
...
+                                                                   mode=UpdateEnums.Modes.AUTOMATIC,
+                                                                   module_address=None,
+                                                                   metadata=metadata)
+                had_work |= successes > 0 or failures > 0
             except Exception as ex:
                 component_logger.error('Could not update {0} to {1}: {2}'.format(firmware_type, target_version, ex))
 
+        if not had_work and self._pending_updates:
+            global_logger.info('No pending updates')
+            self._pending_updates = False
```
ptuytsch, line 533: Correct me if i;m wrong, but I cannot see what is the purpose of this flag? This is never used to check something or if ti continute on updating pending updates?

src/master/core/slave_updater.py
```diff
@@ -138,22 +142,17 @@ def update(address, hex_filename, gen3_firmware, version, logger, slave_communic
...
-                        break
-                    except CommunicationTimedOutException as ex:
-                        logger.warning('Flashing... Block {0} failed: {1}'.format(block, ex))
-                        if tries >= 3:
-                            raise
+                try:
+                    response = slave_communicator.do_command(address=address,
+                                                             command=SlaveAPI.write_firmware_block(),
+                                                             fields={'address': block, 'payload': payload},
+                                                             timeout=SlaveUpdater.WRITE_FLASH_BLOCK_TIMEOUT)
```
ptuytsch, line 149: Is the removal of the 3 retries on purpose here?

src/gateway/update_controller.py
khenderick: The update loop runs every minute, this flag is used to make sure the `No pending updates` is only printed once, and not every minute.

src/master/core/slave_updater.py
khenderick: Yes, every `do_command` on the slave/ucan communicator now tries 3 times by default.

ptuytsch (approved):

PR #924, opened 2021-12-10
eSafe certificate rotation support

ptuytsch:


src/vpn_service.py
```diff
@@ -805,7 +820,10 @@ def run(self, context):
                         self._cloud.authenticate(raise_exception=True)
                         changed = True
                 except Exception:
-                    files.rollback()
+                    try:
```
LnL7, line 823: What's the reason for continuing here if rollback failed anyway?

ptuytsch: edited, also rollback error logging is more verbose now.

LnL7 (approved):

PR #940, opened 2021-12-17
Prevent background update scan

khenderick:


src/gateway/update_controller.py
```diff
@@ -462,6 +463,8 @@ def _execute_pending_updates(self):
         if self._updates_blocked:
             global_logger.info('Updates currently blocked')
             return
+        if not (self._pending_updates or UpdateController.BACKGROUND_UPDATE_SCAN):
```
Shinigami-shinto, line 466: I'm assuming _pending_updates will be true if there is FW to update, in this case when the scan is disbled this condition is only true when there are no pending updates (False)?

Should'nt it be if (pending_updates and scan)?

Shinigami-shinto (approved):

src/gateway/update_controller.py
khenderick: Discussed on slack

PR #944, opened 2021-12-30
BE-131 Gateway thermostat feature on gen3

svanscho:
Altered the get_features API to check which class instances are loaded, instead of the features table. This decision logic is already present in the initialize functions anyhow.

### DIFF
--- a/src/gateway/webservice.py
+++ b/src/gateway/webservice.py
@@ -691,6 +691,10 @@ class WebInterface(object):
             'ventilation',  # Native ventilation
             'modules_information',  # Clean module information
             'dynamic_updates',  # Dynamic updates
+            # TODO: remove
+            'default_timer_disabled',
+            '100_steps_dimmer',
+            'input_states'
         }
         features |= self._module_controller.master_get_features()
         features |= self._thermostat_controller.get_features()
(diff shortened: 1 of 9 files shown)


src/gateway/webservice.py
```diff
@@ -691,27 +691,30 @@ def get_features(self):
...
-            features.append('can_bus_termination_toggle')
+            features.add('can_bus_termination_toggle')
 
-        feature = Feature.get_or_none(name=Feature.THERMOSTATS_GATEWAY)
-        if feature and feature.enabled:
-            features.extend([Feature.THERMOSTATS_GATEWAY, 'thermostat_groups'])
+        if Platform.get_platform() in Platform.ClassicTypes:
+            # TODO: cleanup
+            master_version = self._module_controller.get_master_version()
+            if master_version >= (3, 143, 77):
```
svanscho, line 702: @khenderick Can I assume these are gen2 only features? If so, given they are quite old, can't we just remove them?

LnL7 (approved):

src/gateway/webservice.py
khenderick: While you can indeed remove the check (everybody who will run this code should be running a newer gen2 master anyway), these features should either still be returned (so that the portal does show the new behaviour), or the portal must be updated to always assume this feature is present.

PR #957, opened 2022-01-14
GW logs api

svanscho:


src/gateway/system_controller.py
```diff
@@ -268,6 +268,23 @@ def _restart(_service_names):
...
 
+    def get_logs(self):
+        tmp_dir = tempfile.mkdtemp()
+        try:
+            tmp_log_dir = '{0}/logs'.format(tmp_dir)
+            log_dir = '/var/log/supervisor'
+            shutil.copytree(log_dir, tmp_log_dir)
+
+            archive_filename = 'logs.tar.gz'
+            retcode = subprocess.call('cd {0}; tar czvf {1} *'.format(tmp_log_dir, archive_filename), shell=True)
```
LnL7, line 279: Can't we stream this straight from python using `tarfile`?

svanscho: Not sure we can stream it, but I've at least used an in memory buffer now.

LnL7 (approved):

PR #958, opened 2022-01-19
Return the doorbell icon enable when there are doorbells discovered o…

ptuytsch:
…n the system


src/gateway/system_config_controller.py
```diff
@@ -104,9 +108,14 @@ def save_config_value(cls, config_name, config_value):
...
     @classmethod
-    def get_doorbell_config(cls):
-        # type: () -> SystemDoorbellConfigDTO
+    @Inject
+    def get_doorbell_config(cls, rebus_controller=INJECTED):
+        # type: (RebusController) -> SystemDoorbellConfigDTO
         config = cls.get_config_value('doorbell_config')
+
+        # And the doorbell enable with the fact that there are doorbells present in the rebus system
+        rebus_doorbell_present = len(rebus_controller.get_doorbells()) != 0
```
LnL7, line 117: I would expect the config field to ge updated appropriately by eg. doorbell syncing instead of introducing multiple conditions.

ptuytsch:
This feature will be implemented in the new stripped backend for eSafe

PR #959, opened 2022-01-20
Fixing front panel led control

khenderick:


src/gateway/hal/frontpanel_controller_core.py
```diff
@@ -229,48 +229,87 @@ def _update_cloud_led(self):
...
         else:
-            mode = LedStates.SOLID
-        self._master_controller.drive_led(led=Leds.CLOUD,
-                                          on=on,
-                                          mode=mode)
+            state = LedStates.SOLID
+        self._drive_led(led=Leds.CLOUD, state=state)
+
+    def _drive_led(self, led, state):
+        if led not in FrontpanelCoreController.AVAILABLE_LEDS[self._platform]:
```
LnL7, line 241: Is this an expected case? Otherwise a warning seems appropriate.

src/gateway/hal/frontpanel_controller_core.py
```diff
@@ -92,34 +98,35 @@ def __init__(self, master_communicator=INJECTED):  # type: (CoreCommunicator) ->
...
+                        led = FrontpanelCoreController.LED_MAPPING_ID_TO_ENUM[self._platform][chip].get(led_id)
+                        if led is None:
                             continue
-                        state_tracker = self._led_states.setdefault(led_name, LedStateTracker(led_name))
-                        event_state = core_event.data['leds'].get(led_id, MasterCoreEvent.LedStates.OFF)
-                        state_tracker.set_on(event_state != MasterCoreEvent.LedStates.OFF)
+                        new_state = core_event.data['leds'].get(led_id, MasterCoreEvent.LedStates.OFF)
+                        with self._led_controller_lock:
+                            led_controller = self._led_controllers.setdefault(led, LedController(led))
+                        led_controller.set_on(new_state != MasterCoreEvent.LedStates.OFF)
```
LnL7, line 129: doesn't report changes here?

src/gateway/hal/frontpanel_controller_core.py
khenderick: Yes, based on the firmware docs it seems the Brain doesn't have an LED for the expansion bus, while the Brain+ has one. This way the code that calls `_drive_led` can just drive the expansion bus LED without having to know about the platform. And this method then just discards the request as the LED isn't available, and doesn't need to lock etc.

src/gateway/hal/frontpanel_controller_core.py
khenderick: Correct. The main issue here is that unfortunately the firmware sends 2 separate events when an LED changes state. One for on/off, and one for the blinking mode. This means that I need to process two events, before I can merge this info into a single LED state used elsewhere. The firmware does ensure that the on/off event is send first, and the blinking mode second. This way, this code can always expect to only report changes on the blinking mode (=second) event, reducing intermediate changes.

LnL7 (approved):

PR #1009, opened 2022-02-22
Increase number of tries on uCAN bootload related commands

khenderick:
Increased various tries:
* ping from 2 to 4
* control actions from 3 to 10
* block writes from 6 to 20


svanscho:
Do you some measurements to share that justify the changes? How much better is it?

khenderick:
No. Based on the logs I noticed that most issues are fixed with a single retry, a lot need two retries, some need 3 retries, and so on. However, every now and then the bootloading of a uCAN fails after even 6 retires. @wingels suggested/requested to bump the number of retries to make it more likely that they eventually succeed.

svanscho (approved):

PR #1013, opened 2022-02-23
Fix inconsistent thermostat mode (int vs str)

khenderick:


src/gateway/metrics_collector.py
```diff
@@ -566,7 +566,6 @@ def _run_thermostats(self, metric_type):
                                   'output1': convert_float(thermostat.output_1_level),
                                   'steering_power': convert_float(thermostat.steering_power),
                                   'state': thermostat.state,
-                                  'mode': int(thermostat.mode),
```
LnL7, line 569: Not sure if it matters, but isn't this also defined in the metrics definitions somewhere?

LnL7 (dismissed):

src/gateway/metrics_collector.py
khenderick: Ah, indeed, forgot about those.

LnL7 (approved):

PR #1034, opened 2022-03-09
Add support for master events to control Gateway thermostats

khenderick:


svanscho:
Looks good to me, would just add a unit test for handling the master events as we do in other controllers.

svanscho (changes requested):
Looks good to me, would add a unit test for handling the master events.

svanscho (approved):

PR #1041, opened 2022-03-11
More parsing of query parameters for easier debugging

khenderick:


src/gateway/webservice.py
```diff
@@ -129,19 +129,28 @@ def wrapper(func, *args, **kwargs):
...
-                if mask is None or key not in mask:
-                    params[key] = value
-                else:
-                    params[key] = '***'
-        headers = ' - '.join('{0}: {1}'.format(header, request.headers[header])
-                             for header in ['X-Request-Id', 'User-Agent']
-                             if header in request.headers)
+        headers = ''
+        if access_logger.level == logging.DEBUG:
+            query_string = request.query_string
```
svanscho, line 134: wouldn't it be better to use:
```cherrypy.request.params.get(key_name)```
and hence:
```for q, v in cherrypy.request.params.items()```

svanscho (approved):

PR #1045, opened 2022-03-14
Update output controller

khenderick:


requirements.txt
```diff
@@ -23,7 +23,7 @@ pytz == 2017.3
 apscheduler == 3.6.3
 peewee == 3.13.1
 peewee-migrate == 1.1.6
-sqlalchemy == 1.3.12
+SQLAlchemy==1.4.32
```
svanscho, line 26: perhaps we should use lowercase everywhere?

svanscho (approved):

PR #1046, opened 2022-03-14
Refactored shutter controller

svanscho:


src/gateway/shutter_controller.py
```diff
@@ -193,74 +192,80 @@ def report_shutter_lost_position(self, shutter_id):  # type: (int) -> None
...
-                shutter.save()
-            shutters_to_save.append(shutter_dto)
-        self._master_controller.save_shutters(shutters_to_save)
-        changed = self.update_config(self.load_shutters())
-        if changed:
-            self._publish_config()
+        with Database.get_session() as db:
+            shutters_to_save = []
+            for shutter_dto in shutters:
+                shutter = db.query(Shutter).where(Shutter.number.is_(shutter_dto.id)).one_or_none()
```
LnL7, line 214: I don't think `IS` is the correct operator for number equality.

svanscho: Was a copy from room controller. Changed it everywhere.

LnL7 (approved):

PR #1055, opened 2022-03-17
Convert user controller

khenderick:


src/gateway/user_controller.py
```diff
@@ -193,7 +207,8 @@ def check_token(self, token):
     def generate_new_pin_code(self, length=4):
         # type: (int) -> str
         _ = self
-        current_pin_codes = User.select(User.pin_code).execute()
+        with Database.get_session() as db:
+            current_pin_codes = list(db.query(User.pin_code).all())
```
LnL7, line 211: Does this actually return a list of pin_codes instead of models?

LnL7 (approved):

PR #1121, opened 2022-04-12
Introduce the `in_use` flag

khenderick:


svanscho (dismissed):
Would add a small test for the set_configuration too

ptuytsch (approved):

PR #1129, opened 2022-04-14
Don't turn off/on or toggle lights that are not in use

khenderick:


src/gateway/output_controller.py
```diff
@@ -198,8 +198,9 @@ def save_dimmer_configuration(self, dimmer_configuration_dto):
         self._master_controller.save_dimmer_configuration(dimmer_configuration_dto)
 
     def set_all_lights(self, action):  # type: (Literal['ON', 'OFF', 'TOGGLE']) -> None
-        # TODO: Also include other sources (e.g. plugins) once implemented
-        self._master_controller.set_all_lights(action=action)
+        light_output_ids = [output.id for output in self.load_outputs()
+                            if output.in_use and output.output_type >= 128]
```
ptuytsch, line 202: The output type is a bit redundant here? When/if there will be a new generation of hardware, this check potentially will not match the hardware.

ptuytsch (approved):

src/gateway/output_controller.py
svanscho: @khenderick how much work would it be (now that we have decent migrations) to migrate this to output_type == light first? and then translate the 128 on the master controller level (ORM sync)

svanscho: The API v0 API right now defines an integer (ugly). We could already map the integer to an enum in the DTO, save the proper enum in the ORM and then map it back to some hw specific integer in the master or wherever needed (don't think we actually need it on the gw tbh). The API v1 on the GW (JSON SCHEMA?) might/will/must define the enum directly on the API level then.

PR #1136, opened 2022-04-19
ventilation rooms and registration

LnL7:


src/plugin_runtime/api.py
```diff
@@ -0,0 +1,57 @@
...
+                'plugin': self._plugin_name,
+                'external_id': external_id,
+                'physical_quantity': physical_quantity,
+                'config': config or {}}
+        data = self._api('POST', '/plugin/sensor/register', json=data)
+        return SensorDTO(data['id'], external_id=external_id)
+
+
+class VentilationApi(BaseApi):
+    def register_unit(self, external_id, config=None):
```
svanscho, line 51: would just call this register

testing/unittests/gateway_tests/sensor_controller_test.py
```diff
@@ -288,7 +288,7 @@ def handle_event(gateway_event):
             assert [0] == [x.id for x in master_sensor_dtos]
             assert ['foo'] == [x.name for x in master_sensor_dtos]
 
-    def test_save_sensors_create(self):
+    def test_register_sensor(self):
```
svanscho, line 291: would also add some unit test to safeguard the new API contracts, as there is reuse between plugin and gw APIs

svanscho (approved):

PR #1148, opened 2022-05-06
BE-311 expose ventilation rooms in v0 api for cloud syncing

LnL7:


src/gateway/api/serializers/ventilation.py
```diff
@@ -61,6 +64,8 @@ def deserialize(api_data):  # type: (Dict[str,Any]) -> VentilationDTO
             ventilation_dto.external_id = Toolbox.nonify(api_data['external_id'], '')
         if 'name' in api_data:
             ventilation_dto.name = Toolbox.nonify(api_data['name'], '')
+        if 'room' in api_data:
+            logger.warning('Assigning room not supported in v0 api')
```
svanscho, line 68: I like this

svanscho (approved):

src/gateway/api/serializers/ventilation.py
LnL7: Indeed, those are separate.

PR #1150, opened 2022-05-11
resolve linked shutters based on shutter configuration

LnL7:
This fixes issues when unlinking shutters that are not configured with
the default one-to-one mapping (eg. created with cli)

### DIFF
--- a/src/gateway/hal/master_controller_core.py
+++ b/src/gateway/hal/master_controller_core.py
@@ -684,28 +684,45 @@ class MasterCoreController(MasterController):
         return output_status
 
     def _configure_output_shutter(self, output, is_shutter):  # type: (OutputConfiguration, bool) -> None
-        shutter = ShutterConfiguration(output.id // 2)
-        output_module = output.module
+        output_set = ['01', '23', '45', '67'][output.id % 8 // 2]
+        base_output = output.id // 2 * 2
+        outputs = [OutputConfiguration(base_output),
+                   OutputConfiguration(base_output + 1)]
+        outputs[output.id % 2] = output
+
+        shutter = self._find_output_shutter(base_output)
         if is_shutter:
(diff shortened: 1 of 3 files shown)


src/gateway/hal/master_controller_core.py
```diff
@@ -684,28 +684,45 @@ def load_output_status(self):
...
+    def _find_output_shutter(self, output_id):
+        # type: (int) -> ShutterConfiguration
+        global_configuration = GlobalConfiguration()
+        nr_of_output_modules = global_configuration.number_of_output_modules
+        for module_id in range(nr_of_output_modules):
+            for i in range(4):
+                shutter = ShutterConfiguration(module_id + i)
+                if output_id == shutter.outputs.output_0:
+                    return shutter
+        return ShutterConfiguration(output_id // 2)
```
ptuytsch, line 725: Is there a reason why the shutter config id is chosen to be "mapped" to the output id (shutter_id = output_id // 2). Can this for example not be the next inline that is free?

ptuytsch, line 725: Is there a reason why the shutter config id is chosen to be "mapped" to the output id (shutter_id = output_id // 2). Can this for example not be the next inline that is free?

ptuytsch: furthermore, in theorie you cannot assign all outputs to a shutter config: max output modules = 80, so max outputs = 640 => 320 shutter configs, but there is only place for 256 shutter configs?

ptuytsch: furthermore, in theorie you cannot assign all outputs to a shutter config: max output modules = 80, so max outputs = 640 => 320 shutter configs, but there is only place for 256 shutter configs?

ptuytsch (approved):

PR #1169, opened 2022-05-24
Post-flash boot retries

khenderick:
* Retry post-flash boot
* Few refactorings to speed things up a bit

### DIFF
--- a/src/master/core/core_updater.py
+++ b/src/master/core/core_updater.py
@@ -162,7 +162,7 @@ class CoreUpdater(object):
                     break  # The marker was received
                 except Exception as ex:
                     enter_bootloader_tries -= 1
-                    if enter_bootloader_tries == 0:
+                    if enter_bootloader_tries <= 0:
                         raise
                     self._component_logger.warning(str(ex))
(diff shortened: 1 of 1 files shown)


src/master/core/core_updater.py
```diff
@@ -217,46 +230,58 @@ def update(self, hex_filename, version, ):
...
+                    Hardware.set_gpio(Hardware.CoreGPIO.MASTER_POWER, True)
+                    raise
+                self._component_logger.info('Waiting for startup')
+                if not self._master_started.wait(CoreUpdater.APPLICATION_STARTUP_TIMEOUT):
+                    raise RuntimeError('Core was not started after {0}s'.format(CoreUpdater.APPLICATION_STARTUP_TIMEOUT))
+                self._component_logger.info('Startup complete')
+                break
+            except Exception as ex:
+                post_flash_tries -= 1
+                if post_flash_tries == 0:
```
svanscho, line 266: pretty sure it will work, but as a safety measure I would check counters (decrementing or incrementing) with a >= or <= statement to avoid infinite loops

svanscho (changes requested):

PR #1182, opened 2022-06-15
Screens + Plugins IPC refactor

khenderick:


src/gateway/api/V1/screens.py
```diff
@@ -0,0 +1,95 @@
...
+                                      controller=self, action='list',
+                                      conditions={'method': ['GET']})
+        self.route_dispatcher.connect('sync', '/sync',
+                                      controller=self, action='sync',
+                                      conditions={'method': ['GET']})
+        self.route_dispatcher.connect('retrieve', '/:screen_id',
+                                      controller=self, action='retrieve',
+                                      conditions={'method': ['GET']})
+        for action in ['stop', 'open', 'close']:
+            self.route_dispatcher.connect(action, '/:screen_id/{0}'.format(action),
```
bneyr, line 51: Instead of open & close, Perhaps you could generalise on an action?
/move , args:
action: enum, values (open, close, position, MY_position
value: number, optional, when position is specified

src/gateway/models.py
```diff
@@ -129,6 +129,32 @@ class Sensor(Base):
...
+    type = Column(String(255), nullable=False)
+
+    translational_steps = Column(Integer, nullable=True)
+    rotational_steps = Column(Integer, nullable=True)
+
+    source = Column(String(255), nullable=False)
+    external_id = Column(String(255), nullable=False)
+    plugin_id = Column(Integer, ForeignKey('plugin.id', ondelete='CASCADE'), nullable=True)
+
+    room = relationship('Room', lazy='joined', innerjoin=False)  # type: RelationshipProperty[Optional[Room]]
```
bneyr, line 149: aren't all these (outputs, inputs, sensors, screens) not a common base class (Component) that have a foreign key to a Room? Or even better, a Many2Many between Component & Room?

src/gateway/api/V1/screens.py
khenderick: We'll post-pone this to one of the future tickets where we the actual API implementation/definitions. These calls were merely added to validate the other code.

src/gateway/models.py
khenderick: After some discussions we'll leave this for now. It's a valid comment, but we want to change this together with some more model related cleanups

PR #1217, opened 2022-08-04
simplify plugin startup behaviour

LnL7:
- A plugin will initialise itself automatically when starting, instead
  of waiting for an explicit START action from the gateway service
- The gateway service doesn't wait an extra 30s before registering a
  plugin as ready after it reports it's status after successfully
  sending the START action
- Ensure that the metrics poller only starts once the gateway service
  has retrieved metric definitions, otherwise incorrect error messages
  are logged during startup
- Initialize connectors in main thread of plugins to ensure that the
  metrics poller doesn't exit as soon as it's parent IPC request thread
  exits


src/gateway/webservice.py
```diff
@@ -2658,26 +2658,8 @@ def stop(self):
...
-        try:
-            self._http_server.stop()
-        except Exception as ex:
-            logger.error('Could not stop non-secure webserver: {0}'.format(ex))
-        try:
-            self._https_server.stop()
-        except Exception as ex:
-            logger.error('Could not stop secure webserver: {0}'.format(ex))
         try:
             for mount in mounts:
```
svanscho, line 2662: Does this also include unmounting e.g. when a plugin stops?
Probably not, but wasn't the case before either I guess.

src/plugin_runtime/connectors/metrics.py
```diff
@@ -100,7 +98,8 @@ def _poll_metrics(self):
         now = time.time()
         for handler_name, handler_info in self._polling_handlers.items():
             handler, interval = handler_info
-            last_poll = self._last_poll.get(handler_name, 0.0)
+            last_poll = self._last_poll.get(handler_name, now + 60)  # delay at startup
```
svanscho, line 101: Not sure if now and the embedded context justify, but using some timedelta classes over time might improve readability.

src/plugin_runtime/runtime.py
```diff
@@ -215,6 +206,19 @@ def _start_background_tasks(self):
...
+        # type: () -> None
+        if self._metric_definitions:
+            used_decorators = set()
+            for decorated_method, _ in get_special_methods(self._plugin, 'om_metric_data'):
+                used_decorators.add('om_metric_data')
+                self._connector.metrics._attach_collector(handler=decorated_method,
+                                                          interval=decorated_method.om_metric_data['interval'])
+            # TODO: Deprecate decorators in favor of connectors when needed
+            # for decorator_name in used_decorators:
+            #     global_logger.warning('The `{0}` decorator is deprecated. Use the connector instead'.format(decorator_name))
```
svanscho, line 219: Good idea. Why not enable it already and change message to "will be deprecated in a future version"

svanscho (approved):

PR #1253, opened 2022-09-08
dummy improvements

LnL7:


src/gateway/hal/master_controller_core_dummy.py
```diff
@@ -224,6 +231,32 @@ def do_command(self, command, fields, timeout=2, bypass_blockers=None):
                     'timer_type': output_state.get('timer_type', 0), 'timer_type_standard': 0,
                     'timer': output_state.get('timer', 0), 'timer_standard': 0,
                     'group_action': 255, 'dali_output': 255, 'output_lock': 0}
+        if command.instruction == b'SI':
+            global_configuration = GlobalConfiguration()
+            values = [None] * 8 * global_configuration.number_of_sensor_modules  # type: List[Any]
```
ptuytsch, line 236: this indicates that there will be 8 * number of sensor modules return values, but there are only 8 returned upon request (the complete set per module)

as far as I know

ptuytsch (approved):

PR #1260, opened 2022-10-04
Add plugin logging plus fix double plugin startup

ptuytsch:


src/gateway/sensor_controller.py
```diff
@@ -88,6 +89,7 @@ def _plugin_callback_action(self, source, action, action_payload):
                 self._publish_config()
             return sensor_dto
         if action == SensorEnums.CallbackAction.REPORT_STATUS:
+            logger.debug('Sensor registers status based on payload: {}'.format(action_payload))
```
LnL7, line 92: Using "register" here is somewhat confusing

src/gateway/watchdog.py
```diff
@@ -101,6 +101,7 @@ def _controller_health(self, name, controller, device_reset):
                 device_reset()
                 if reset_action == 'service':
                     self._update_controller.block_updates()
+                    logger.error('Master communication issues, restarting service...')
```
LnL7, line 104: This doesn't only trigger for master communication, the name/controller can be either master or energy.

src/plugin_runtime/runtime.py
```diff
@@ -33,6 +34,33 @@
...
+        if self.file_logger is None and self.enabled:
+            self.file_logger = open(filename, 'a')
+
+    def log(self, obj, msg):
+        if self.enabled and self.file_logger is not None:
+            self.file_logger.write("{} [runtime: {}]: {}\n".format(datetime.datetime.now().replace(microsecond=0).isoformat(), obj, msg))
+            self.file_logger.flush()
+
+
+runtime_logger = RuntimeLogger(enabled=False)
```
LnL7, line 54: What is this for? Looks like functionality that's covered by the native logging library.

src/gateway/watchdog.py
ptuytsch: This was indeed to debug another issue, Was not needed in this context. Is removed.

src/logs.py
```diff
@@ -14,13 +14,15 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import logging
-import sys
 from logging import handlers, Handler
 import os
 import re
-import constants
+from six.moves.configparser import ConfigParser, NoOptionError, NoSectionError
```
LnL7, line 20: With the python2 environment 3rd party modules won't be available until `System.import_libs()` is called. Do all the entrypoints do this before anything related to logging is setup and imported?

LnL7 (approved):

src/logs.py
ptuytsch: Had to check, but yes this is one of the first things that happens on the plugin runtine as well. This does also work on a python2 running gateway.

PR #1267, opened 2022-10-17
valve-pump fixes

hendrikVandeput:


src/openmotics_service.py
```diff
@@ -238,6 +240,8 @@ def start(master_controller=INJECTED,  # type: MasterController
         pubsub.start()
         system_controller.start()
         update_controller.start()
+        valve_pump_controller.start()
```
svanscho, line 243: wouldn't it make more sense to start this from within the thermostat controller? Not sure why this would be needed when not using thermostats...

svanscho (changes requested):

src/openmotics_service.py
hendrikVandeput: valid point, updated

PR #1273, opened 2022-10-25
pump group fix

hendrikVandeput:


src/gateway/api/serializers/thermostat.py
```diff
@@ -197,16 +197,20 @@ def serialize(pump_group_dto, fields):  # type: (PumpGroupDTO, Optional[List[str
...
+        pump_group_dto = PumpGroupDTO(id=api_data['id'])
         SerializerToolbox.deserialize(
             dto=pump_group_dto,  # Referenced
             api_data=api_data,
             mapping={'output': ('pump_output_id', PumpGroupSerializer.BYTE_MAX),
-                     'rooom': ('room_id', PumpGroupSerializer.BYTE_MAX)}
+                     'room': ('room_id', PumpGroupSerializer.BYTE_MAX)}
         )
+        if 'output' in api_data:
+            pump_group_dto.pump_output_id = api_data['output']
```
LnL7, line 208: In general the v0 uses 0-254 as valid number identifiers and 255 as null. The rest of the code assumes that the later gets translated to None. This is what the defined mappings above already do so explicitly handling the `output` and `room` fields isn't necessary.

testing/unittests/gateway_tests/webservice_test.py
```diff
@@ -332,7 +332,7 @@ def test_set_pump_group_configurations(self):
             ])
             save.assert_called_with([
                 PumpGroupDTO(0, pump_output_id=1, valve_output_ids=[8, 9, 10]),
-                PumpGroupDTO(0, pump_output_id=None, valve_output_ids=[])
+                PumpGroupDTO(1, pump_output_id=255, valve_output_ids=[])
```
LnL7, line 335: Apart from the 0-4 number of the pump group, what this test tried to validate was correct and should still be translated as None.

LnL7 (changes requested):

LnL7 (approved):

PR #1274, opened 2022-10-27
introduce cloud log events for debugging purposes

LnL7:


svanscho (approved):

svanscho:
> 

We could also add a feature flag and then hide/read only the flag in the cloud. We are syncing the feature flags anyway now.

svanscho (approved):

PR #1283, opened 2022-11-10
remap front-end value to master value and vice-versa

hendrikVandeput:


Shinigami-shinto (approved):

src/gateway/hal/mappers_classic/thermostat.py
```diff
@@ -94,7 +94,15 @@ def orm_to_dto(orm_object):  # type: (EepromModel) -> ThermostatGroupDTO
                 value_field = 'switch_to_{0}_value_{1}'.format(mode, i)
                 dto_field = 'switch_to_{0}_{1}'.format(mode, i)
                 output = Toolbox.nonify(data[output_field], ThermostatGroupMapper.BYTE_MAX)
-                value = Toolbox.nonify(data[value_field], ThermostatGroupMapper.BYTE_MAX)
+
+                # remap 0-63,255 (master) value to 0-100,None (front-end)
+                temp_value = data[value_field]
```
svanscho, line 99: are we sure these mappers are not used when using gen2 (classic) in combination with the "gateway thermostats" feature?

svanscho (changes requested):

src/gateway/hal/mappers_classic/thermostat.py
hendrikVandeput: this mapper is only used in `src/gateway/hal/master_controller_classic.py`

svanscho (approved):

PR #1284, opened 2022-11-14
Feature/module serial number

ptuytsch:
This implements the syncing and exposing of the serial number of an power/energy module over the api.

Also, some api hardening is done to show exceptions that occur during the serialization of the api response.


svanscho (approved):

testing/unittests/gateway_tests/module_controller_test.py
```diff
@@ -151,25 +151,29 @@ def test_serialization(self):
                                   address='2',
```
svanscho, line 151: perhaps add a test for 0000000 and 12345678 ?

svanscho (changes requested):

svanscho (approved):

PR #1294, opened 2022-12-07
OME-426-api-tests-for-gw-integration-testing

felix-capon:


testing/cicd/tests/api_tests/__init__.py
```diff
@@ -0,0 +1 @@
+# no need for this
```
ptuytsch, line 1: just add the copyright header here, and indeed leave it empty :)

ptuytsch (approved):

PR #1297, opened 2022-12-15
OME-389 Feature/energy discovery

ptuytsch:


src/gateway/energy/energy_communicator.py
```diff
@@ -137,347 +156,326 @@ def get_seconds_since_last_success(self):
...
+    def _read_packets(self):
+        packet = EnergyPacket()
+
+        while not self._stop:
+            byte = None
+            try:
+                if not self._serial_port.data_available():
+                    time.sleep(0.1)
+                    continue
+                byte = self._serial_port.read(1)
```
bneyr, line 272: I think all of this works, but I wonder if it is super efficient to read the bytes one by one.
usually in these implementations, you perform a blocking read for the first character, but after that you read all the in_waiting bytes at once.

src/gateway/energy/energy_packet.py
```diff
@@ -0,0 +1,399 @@
...
+        self.__mode = None  # type: Optional[bytearray]
+        self.type = bytearray()
+        self.__length = None  # type: Optional[bytearray]
+        self.data = bytearray()
+        self.__crc = None  # type: Optional[bytearray]
+
+        self.complete = False
+        self._append_status = None
+        self._append_helpers = {
+            "start_1": EnergyPacketReceiveHelper('is_request', lambda p, x: 'start_2' if x in [b'S', b'R'] else EnergyPacketReceiveHelper.Status.ERROR, 'Invalid packet start sequence'),
```
bneyr, line 45: Wouldn't it be better to declare StrEnums for these as well? One type in this cryptic block and the whole state machine is broken
Another way to clean this up a bit is to use a real inheritance-based state machine, which would improve the readability a lot: https://python-3-patterns-idioms-test.readthedocs.io/en/latest/StateMachine.html

src/gateway/energy/energy_packet.py
```diff
@@ -0,0 +1,399 @@
...
+            "start_2": EnergyPacketReceiveHelper(None, lambda p, x: 'start_3' if x in [b'T'] else EnergyPacketReceiveHelper.Status.ERROR, 'Invalid packet start sequence'),
+            "start_3": EnergyPacketReceiveHelper(None, lambda p, x: 'address_type' if x in [b'R'] else EnergyPacketReceiveHelper.Status.ERROR, 'Invalid packet start sequence'),
+            "address_type": EnergyPacketReceiveHelper('address_type', lambda p, x: 'address_id' if x in [b'E', b'H', b'C'] else EnergyPacketReceiveHelper.Status.ERROR, 'Invalid module type'),
```
bneyr, line 48: Please expand the module type enum and just check if the address_id is a member of that enum.

src/gateway/energy/energy_packet.py
```diff
@@ -0,0 +1,399 @@
...
+            "address_id": EnergyPacketReceiveHelper('address_id', lambda p, x: 'command_id' if Toolbox.is_int(ord(x)) else EnergyPacketReceiveHelper.Status.ERROR, 'Invalid module id'),
```
bneyr, line 49: is_int on an "ord" function of a byte is not that useful I think.
The previous validator already covers this, no?

src/gateway/energy/energy_packet.py
```diff
@@ -0,0 +1,399 @@
...
+            "command_id": EnergyPacketReceiveHelper('command_id', lambda p, x: 'mode' if Toolbox.is_int(ord(x)) else EnergyPacketReceiveHelper.Status.ERROR, 'Invalid command id'),
```
bneyr, line 50: Same argument. A byte is always 0-255, all values are valid, so a validator is not needed imo

src/gateway/energy/energy_packet.py
```diff
@@ -0,0 +1,399 @@
...
+            "mode": EnergyPacketReceiveHelper('mode', lambda p, x: 'type' if x in map(lambda c: c.value, EnergyCommandMode) else EnergyPacketReceiveHelper.Status.ERROR, 'Invalid command mode'),
+            "type": EnergyPacketReceiveHelper('type', lambda p, x: 'length' if len(p.type) >= 2 else EnergyPacketReceiveHelper.Status.CONTINUE, 'Invalid command type'),
+            "length": EnergyPacketReceiveHelper('length', lambda p, x: EnergyPacketReceiveHelper.Status.ERROR if not Toolbox.is_int(ord(x)) else 'data' if ord(x) > 0 else 'crc', 'Invalid command length'),
```
bneyr, line 53: Personally, I don't think this a readable line.

src/gateway/energy/energy_packet.py
```diff
@@ -0,0 +1,399 @@
...
+            "start_3": EnergyPacketReceiveHelper(None, lambda p, x: 'address_type' if x in [b'R'] else EnergyPacketReceiveHelper.Status.ERROR, 'Invalid packet start sequence'),
+            "address_type": EnergyPacketReceiveHelper('address_type', lambda p, x: 'address_id' if x in [b'E', b'H', b'C'] else EnergyPacketReceiveHelper.Status.ERROR, 'Invalid module type'),
+            "address_id": EnergyPacketReceiveHelper('address_id', lambda p, x: 'command_id' if Toolbox.is_int(ord(x)) else EnergyPacketReceiveHelper.Status.ERROR, 'Invalid module id'),
+            "command_id": EnergyPacketReceiveHelper('command_id', lambda p, x: 'mode' if Toolbox.is_int(ord(x)) else EnergyPacketReceiveHelper.Status.ERROR, 'Invalid command id'),
+            "mode": EnergyPacketReceiveHelper('mode', lambda p, x: 'type' if x in map(lambda c: c.value, EnergyCommandMode) else EnergyPacketReceiveHelper.Status.ERROR, 'Invalid command mode'),
+            "type": EnergyPacketReceiveHelper('type', lambda p, x: 'length' if len(p.type) >= 2 else EnergyPacketReceiveHelper.Status.CONTINUE, 'Invalid command type'),
```
bneyr, line 52: isn't type always 3 bytes?

src/gateway/energy/energy_packet.py
```diff
@@ -0,0 +1,399 @@
...
+        # type: () -> bytearray
+        self._populate_raw_data()
+        crc = self.calculate_crc()
+        self.crc = crc
+        self.raw_data += bytearray([crc]) + CRNL
+        self.complete = True
+        return self.raw_data
+
+    def append_byte(self, byte):
+        # type: (bytearray) -> bool
```
bneyr, line 241: For me this is a bit confusing. The function is called "append_byte", however the argument is an array of bytes.

src/gateway/energy/energy_packet.py
```diff
@@ -0,0 +1,399 @@
...
+        result = self._append_helpers[self._append_status].validate(self, byte)
+        if result in EnergyPacketReceiveHelper.Status.ALL:
+            if result == EnergyPacketReceiveHelper.Status.CONTINUE:
+                self.raw_data += byte
+            elif result == EnergyPacketReceiveHelper.Status.ERROR:
+                self.reset()
+            elif result == EnergyPacketReceiveHelper.Status.END:
+                self.raw_data += byte
+                self.complete = True
+        elif isinstance(result, str):
```
bneyr, line 260: Just use enums please

src/gateway/enums.py
```diff
@@ -301,6 +301,8 @@ class Version(object):
     VERSION_TO_STRING = {Version.POWER_MODULE: 'power',
                          Version.ENERGY_MODULE: 'energy',
                          Version.P1_CONCENTRATOR: 'p1_concentrator'}
+    CHAR_TO_VERSION = {'E': Version.ENERGY_MODULE,
```
bneyr, line 304: Can't we just make Version an enum with some helper functions?
e.g;

```
Class Version(str, Enum):
    POWER_MODULE = 'power'
    
    @classmethod
    def from_module_type(cls, module_type):
        if module_type =  EnergyModuleType.P:
            return cls.POWER_MODULE
        
```

src/gateway/energy/energy_api.py
```diff
@@ -248,9 +248,9 @@ def get_day_energy(version):
         :param version: power api version (POWER_API_8_PORTS or POWER_API_12_PORTS).
         """
         if version == EnergyEnums.Version.POWER_MODULE:
-            return EnergyCommand('G', 'EDA', '', '8L')
+            return EnergyCommand('G', 'EDA', '', '=8L')
```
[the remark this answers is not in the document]
ptuytsch: indeed good catch, The reason I've added this was for testing it on my local machine. But adding the extra saftey for endianness is also a valid point. I've used the `!` to indicate network endianness.

src/gateway/energy/energy_bus_controller.py
```diff
@@ -0,0 +1,451 @@
...
+                response = context.communicator._do_command(BROADCAST_ADDRESS, cmd, 1, bytearray(module_serial_number, encoding='ascii'), seed, new_address)
+                if response is not None:
+                    module_type = ModuleType.P1_CONCENTRATOR if module_version == 'C' else ModuleType.POWER
+                    if not existing:
+                        EnergyBusDiscoveryMode._save_new_module(new_address, module_type, EnergyEnums.CHAR_TO_VERSION[module_version], module_serial_number)
+            except EnergyException:
+                logger.error('Could not set new address for module with serial number: {}'.format(module_serial_number))
+
+    @staticmethod
+    def _save_new_module(new_address, module_type, module_version, module_serial_number=None):
```
LnL7, line 374: This looks like duplicated code between address and discovery mode, move to a or static helper function.

src/gateway/energy/energy_bus_controller.py
```diff
@@ -0,0 +1,451 @@
...
+                address = int(module.address)
+
+            return address, module is not None
+
+    @staticmethod
+    def receive_packet(context, energy_packet):
+        # type: (EnergyBusContext, EnergyPacket) -> None
+        logger.debug("Received an packet in discovery mode: {}".format(energy_packet))
+        if energy_packet.type not in ['ADD']:
+            logger.error('Energy Bus: Received unexpected packet in addressing mode: {}'.format(energy_packet.printable_data()))
```
LnL7, line 427: Is this completely unexpected, not all modules will support discovery mode.

src/gateway/energy/energy_bus_controller.py
```diff
@@ -0,0 +1,451 @@
...
+        if not isinstance(serial_number, str):
+            logger.error('Received Get ADDress packet, but no valid data in packet: {}'.format(energy_packet))
+            return
+        if module_version is None:
+            logger.error('Received Get ADDress packet, address type is None: {}'.format(energy_packet))
+            return
+
+        logger.debug("Received discovery packet from energy bus from serial: {}; {}".format(serial_number, random_seed))
+        data = (random_seed, module_version)
+        if serial_number in context.discovery_buffer:
```
LnL7, line 448: This is a bit more concise.

```suggestion
        context.discovery_buffer.setdefault(serial_number, [])
        context.discovery_buffer[serial_number].append(data)
```

src/gateway/energy/energy_bus_controller.py
```diff
@@ -0,0 +1,451 @@
...
+        # type: (EnergyBusContext, int, EnergyCommand, *Any) -> Optional[EnergyCommand]
+        raise CommunicationBusUnavailable('Energy bus is in discovery mode.')
+
+    @staticmethod
+    def _distribute_addresses(context):
+        for module_serial_number in context.discovery_buffer.keys():
+            seeds = list(set(context.discovery_buffer[module_serial_number]))  # type: List[Tuple[int, str]]
+            if len(seeds) != 1:
+                logger.error('Not distributing address for module with serial number: {}, multiple seeds are received'.format(module_serial_number))
+            seed, module_version = seeds[0]
```
LnL7, line 360: I find *seed* to be a bit confusing in this context. It makes me expect that the data is used as random input for the address, which is not the case. Something like nonce might be a better name.

src/gateway/energy/energy_communicator.py
```diff
@@ -137,347 +156,323 @@ def get_seconds_since_last_success(self):
...
-    def __get_cid(self):
-        # type: () -> int
-        """ Get a communication id """
-        (ret, self.__cid) = (self.__cid, (self.__cid % 255) + 1)
-        return ret
+class EnergyConsumerPool(object):
+    def __init__(self):
+        # list of command id's with their consumer
+        self.consumers = {}  # type: Dict[int, EnergyConsumer]
+        self._next_cid = 0
```
LnL7, line 164: While probably not a problem in practice, does *all* communication use the CID from a request, including the spontaneous messages in address mode? If not it's probably better to reserve some ids for these type of special messages.

src/toolbox.py
```diff
@@ -257,6 +257,14 @@ def _process_request(self, ipc_dto):
 
 
 class Toolbox(object):
+    @staticmethod
+    def is_int(value):
+        try:
```
LnL7, line 262:
```suggestion
        return str.isdigit(value)
```

src/serial_utils.py
```diff
@@ -75,44 +75,22 @@ def __init__(self, serial):
         """ Initialize a rs485 connection using the serial port. """
         self._serial = serial
         fileno = serial.fileno()
-        if fileno is not None:
+        if fileno is not None and Platform.get_platform() not in Platform.DummyTypes:
```
LnL7, line 78: Is there any reason to keep the RS485 class around instead of moving this startup logic to the energy communicator?

FYI the code for the uart controller also includes logic to switch to rs485 mode, perhaps the code can be shared.

src/gateway/energy/energy_communicator.py
```diff
@@ -137,347 +156,323 @@ def get_seconds_since_last_success(self):
...
+            logger.warning('Received an energy response but no consumer is registered for this request: {}'.format(energy_packet))
+            self._notify_background_consumers(energy_packet)
+        else:
+            consumer.consume(energy_packet)
+
+    def _read_packets(self):
+        while not self._stop:
+            byte = None
+            try:
+                if self._serial_port.inWaiting() == 0:
```
LnL7, line 265: Since the code processes byte by byte anyway and the serial has a 2s timeout, I don't think there's any reason for manual waiting and sleeping instead of just leveraging the read timeout.

src/openmotics_service.py
```diff
@@ -202,8 +204,7 @@ def start(master_controller=INJECTED,  # type: MasterController
 
         # Start rest of the stack
         maintenance_controller.start()
-        if energy_serial:
-            energy_serial.start()
+        energy_communicator.start()
```
LnL7, line 207: The energy_communicator can be `None` if the serial port is disabled (eg. gen2 where the energy bus is used for another integration).

src/gateway/energy/energy_communicator.py
ptuytsch: This is indeed the case, the line just below defines the range of CID's that can be taken for an valid message. The CID will always be checked against that range. The next_cid variable has been renamed for a more describing name to make this clear.

LnL7 (approved):

PR #1311, opened 2023-01-04
Input tests marked unstable

felix-capon:


hendrikVandeput (commented):
Why are both tests moved to unstable?

felix-capon:
Both tests seem to be the only ones that fail 
image

hendrikVandeput:
> Both tests seem to be the only ones that fail image

Why are they failing and why should they be allowed to fail?
one of them is a smoke test so I'm expecting the test to be important

felix-capon:
> > Both tests seem to be the only ones that fail image
> 
> Why are they failing and why should they be allowed to fail? one of them is a smoke test so I'm expecting the test to be important

Tests failure reasons are unkown, currently we want the nightly report to work thus surpassing unstable tests.

hendrikVandeput (dismissed):

hendrikVandeput (approved):

PR #1329, opened 2023-01-16
Change default shutter timer value to 45 seconds

ptuytsch:


LnL7 (commented):
I would expect the portal to always send full configuration payloads, so as a result this fallback won't ever be used.

PR #1331, opened 2023-01-18
Bugfix/ventilation level

felix-capon:


hendrikVandeput (approved):

src/gateway/api/serializers/ventilation.py
```diff
@@ -38,12 +38,15 @@ class VentilationSerializer(object):
...
     def serialize(ventilation_dto, fields):
         # type: (VentilationDTO, Optional[List[str]]) -> Dict[str,Any]
+        amount_of_levels = Toolbox.denonify(ventilation_dto.amount_of_levels, 1)  # seperated for validation of min-max
         data = {'id': ventilation_dto.id,
                 'external_id': Toolbox.denonify(ventilation_dto.external_id, ''),
                 'source': SourceSerializer.serialize(ventilation_dto.source, fields=None),
                 'name': Toolbox.denonify(ventilation_dto.name, ''),
                 'room': Toolbox.denonify(ventilation_dto.room, VentilationSerializer.BYTE_MAX),
-                'amount_of_levels': Toolbox.denonify(ventilation_dto.amount_of_levels, 0),
+                'amount_of_levels': amount_of_levels,
```
svanscho, line 47: is this a deliberate change?

src/plugin_runtime/connectors/ventilation.py
```diff
@@ -58,11 +58,13 @@ def _process_action(self, action, payload, handler, version):
 
     # Driver (e.g. a plugin that integrates a thrid-party ventilation system)
 
-    def register(self, external_id, name, amount_of_levels, device_type, device_vendor, device_serial):
+    def register(self, external_id, name, amount_of_levels, min_level, max_level, device_type, device_vendor, device_serial):
```
svanscho, line 61: I'd expect the min_level and max_level as optional kwargs at the end of the signature for backwards compatibility.
Right now, older plugins will not be able to register ventilation units any longer. I thought the whole purposes was to extend with backwards compatibility.

```def register(self, external_id, name, amount_of_levels, device_type, device_vendor, device_serial, min_level=None, max_level=None ):```

File: src/gateway/models.py [hunk 3]
```diff
@@ -372,6 +372,8 @@ class Sources(object):
     name = Column(String(255), nullable=False)
     room_id = Column(Integer, ForeignKey('room.id', ondelete='SET NULL'), nullable=True)
     amount_of_levels = Column(Integer, nullable=False)
+    min_level = Column(Integer, nullable=False, default=0)
```
svanscho, line 375: why default 0 and with nullable False? I'd expect these values can be None, meaning they are not defined and only the amount_of_levels will be taken into account.

src/gateway/dto/ventilation.py
```diff
@@ -26,30 +26,49 @@
...
 
+        min_level, max_level, amount_of_levels = VentilationDTO.validate_levels(min_level, max_level, amount_of_levels)
+        self.amount_of_levels = amount_of_levels # type: int
+        self.min_level = min_level  # type: int
+        self.max_level = max_level  # type: int
+
+    @staticmethod
+    def validate_levels(min_level, max_level, amount_of_levels):
+        if amount_of_levels > 1:
+            if max_level == 0:
```
svanscho, line 49: would expect if max_level is None here

src/gateway/dto/ventilation.py
```diff
@@ -26,30 +26,49 @@
...
+                max_level = amount_of_levels-1
+        else:
+            amount_of_levels = 1
+            min_level = 0
```
svanscho, line 53: and both None here too

src/gateway/api/serializers/ventilation.py
```diff
@@ -38,12 +38,15 @@ class VentilationSerializer(object):
...
+                'min_level': Toolbox.denonify(ventilation_dto.min_level, 0),
```
svanscho, line 48: Not sure why None(s) are not allowed (see below).

svanscho (changes requested):

svanscho:
@felix-capon few comments, but I'll let @LnL7 have a look too. He is more into the JSON schema things.

Regardless of my comments, can you squash the commits for a more cleaner merge?

src/gateway/api/serializers/ventilation.py
```diff
@@ -67,6 +70,10 @@ def deserialize(api_data):  # type: (Dict[str,Any]) -> VentilationDTO
             logger.warning('Assigning room not supported in v0 api')
         if 'amount_of_levels' in api_data:
             ventilation_dto.amount_of_levels = Toolbox.nonify(api_data['amount_of_levels'], '')
+        if 'min_level' in api_data:
+            ventilation_dto.min_level = Toolbox.nonify(api_data['min_level'], '')
```
LnL7, line 74: The second argument of notify/denonify is used for conversion of None values in dto fields used throughout the code internally. This is because historically many of the configuration apis on the gateway used 255 instead of the json `null`. So `""` should definitely not be used for fields that are expected to container integers.

```
Toolbox.deserializer({"name": ""}) => VentilationDTO(name=None)
Toolbox.serialize(VentilationDTO(name=None)) => {"name":""}
```

The exact values used in this serializer are only used by cloud syncing IIRC, everything in the frontend uses the new `/api/ventilation/units` endpoint with a schema which has a more flexible approach for changes.

src/gateway/api/serializers/ventilation.py
felix-capon: The denonification is done a few lines above. This is because the max amount needs this value for denonification.

src/plugin_runtime/connectors/ventilation.py
felix-capon: Applied

src/gateway/dto/ventilation.py
felix-capon: Applied

src/gateway/api/serializers/ventilation.py
felix-capon: Applied

src/gateway/dto/ventilation.py
```diff
@@ -26,30 +26,49 @@
...
+            max_level = 0
+        return min_level, max_level, amount_of_levels
+
 
 class PluginVentilationDTO(BaseDTO):
-    def __init__(self, name, amount_of_levels, device_type, device_vendor, device_serial, id=None, external_id=None):
+    def __init__(self, name, device_type, device_vendor, device_serial, amount_of_levels, min_level=0, max_level=0, id=None, external_id=None):
```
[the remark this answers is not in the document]
felix-capon: Min and max are now optional.

src/gateway/api/serializers/ventilation.py
felix-capon: Min and max can now be left None (Validation happens later). Amount of levels is denonified to integer.

src/gateway/dto/ventilation.py
```diff
@@ -26,30 +26,49 @@
...
         self.device_type = device_type  # type: str
         self.device_serial = device_serial  # type: str
 
+        min_level, max_level, amount_of_levels = VentilationDTO.validate_levels(min_level, max_level, amount_of_levels)
+        self.amount_of_levels = amount_of_levels # type: int
+        self.min_level = min_level  # type: int
+        self.max_level = max_level  # type: int
+
+    @staticmethod
+    def validate_levels(min_level, max_level, amount_of_levels):
```
svanscho, line 47: perhaps we should declare this outside of the dtos as a helper function?

src/gateway/models.py
```diff
@@ -372,6 +372,8 @@ class Sources(object):
...
+    max_level = Column(Integer, nullable=False, default=0)
```
svanscho, line 376: remove the defaults

src/gateway/dto/ventilation.py
svanscho: we should raise invalid combinations e.g. max_level < min_level

testing/unittests/gateway_tests/webservice_test.py
```diff
@@ -381,7 +383,9 @@ def test_set_ventilation_configuration(self):
             response = self.web.set_ventilation_configuration(config=config)
             self.assertEqual({
                 'id': ventilation_id,
-                'amount_of_levels': 0,
+                'amount_of_levels': 1,
+                'min_level': 0,
```
svanscho, line 387: shouldn't this raise an error?

Would also add a test with amount_of_levels > 1

testing/unittests/plugins_tests/connector_test.py
```diff
@@ -500,10 +500,12 @@ def set_manual(self, external_id, level, timer):
         return 'set_manual_processed'
 
     @om_expose(auth=False)
-    def register(self, external_id, name, amount_of_levels, device_type, device_vendor, device_serial):
+    def register(self, external_id, name, amount_of_levels, min_level, max_level, device_type, device_vendor, device_serial):
```
svanscho, line 503: add None kwargs

svanscho (changes requested):

src/gateway/dto/ventilation.py
svanscho:
```
if min is None or max is None:
    if amount_of_levels == 1:
         min, max = 1, 1
    if amount_of_levels >= 2:
          min, max = 1, amount_of_levels 
 if max < min or amount_of_levels < 1:
     raise ValueError
  ```

**Review by svanscho — CHANGES_REQUESTED** [review]

File: src/gateway/api/serializers/ventilation.py [hunk 13]
```diff
@@ -67,6 +70,10 @@ def deserialize(api_data):  # type: (Dict[str,Any]) -> VentilationDTO
             logger.warning('Assigning room not supported in v0 api')
         if 'amount_of_levels' in api_data:
             ventilation_dto.amount_of_levels = Toolbox.nonify(api_data['amount_of_levels'], '')
```
LnL7, line 72:
```suggestion
            ventilation_dto.amount_of_levels = api_data['amount_of_levels']
```

src/gateway/api/serializers/ventilation.py
LnL7, line 74:
```suggestion
            ventilation_dto.min_level = api_data['min_level']
```

src/gateway/api/serializers/ventilation.py
```diff
@@ -67,6 +70,10 @@ def deserialize(api_data):  # type: (Dict[str,Any]) -> VentilationDTO
...
+        if 'min_level' in api_data:
+            ventilation_dto.min_level = Toolbox.nonify(api_data['min_level'], '')
+        if 'max_level' in api_data:
+            ventilation_dto.max_level = Toolbox.nonify(api_data['max_level'], '')
```
LnL7, line 76:
```suggestion
            ventilation_dto.max_level = api_data['max_level']
```

LnL7 (dismissed):

LnL7 (approved):

PR #1332, opened 2023-01-24
Fix unnecessary deprecation warning

Shinigami-shinto:
The Syncer plugin had a deprecation warning but looking at the arguments, only None was contained, so getting this warning didn't make sense.


Shinigami-shinto:
Context:
https://github.com/openmotics/plugins/pull/88
and Slack discussion

PR #1343, opened 2023-02-13
import protobuf contracts from renson-embedded-contracts package

bneyr:


src/gateway/zmq/zmq_proto_contracts.py
```diff
@@ -0,0 +1,36 @@
+import sys
+
+"""
+Small stub to make sure nothing accidently imports protobuf when running in Py2
+This can be removed once Python2 runtime is fully deprecated
+"""
+
+if sys.version_info >= (3, 0):
```
LnL7, line 8: Since this logic exists anyway, perhaps `ImportError` is a better condition?

LnL7 (approved):

PR #1355, opened 2023-02-24
Hot water

hendrikVandeput:


src/gateway/api/V1/serializers/hot_water.py
```diff
@@ -0,0 +1,29 @@
...
+if False:  # MYPY
+    from typing import Any, Dict, List, Optional
+
+class HotWaterSerializer(object):
+
+    @staticmethod
+    def serialize(hot_water_status_dto):
+        # type: (HotWaterStatusDTO) -> Dict[str,Any]
+        data = {'id': hot_water_status_dto.id,
+                'state': hot_water_status_dto.state,
```
LnL7, line 13: These v1 api calls are used by the forms rendered based on jsonschema so we need to define the schema first.

src/gateway/hot_water_controller.py
```diff
@@ -0,0 +1,211 @@
...
+                    )
+                    db.add(hot_water_orm)
+                    db.commit()
+                logger.info("new hot water unit registered with external_id {}".format(action_payload.external_id))
+                return PluginHotWaterDTO(
+                    id=hot_water_orm.id,
+                    name=hot_water_orm.name,
+                    external_id=hot_water_orm.external_id,
+                )
+        if action == HotWaterEnums.CallbackAction.REPORT_MODE:
```
LnL7, line 92: If we want to support partial status updates I would implement this using `status_dto.loaded_fields` instead of separate actions for each status field.

src/gateway/api/V1/serializers/hot_water.py
```diff
@@ -0,0 +1,51 @@
...
+        data = {'id': hot_water_dto.id,
+                'external_id': hot_water_dto.external_id,
+                'name': hot_water_dto.name,
+                }
+        if hot_water_dto.plugin:
+            data['plugin'] = hot_water_dto.plugin.name
+        if hot_water_dto.source:
+            data['source'] = hot_water_dto.source.source
+        if hot_water_dto.status:
+            data['status'] = hot_water_dto.status
```
LnL7, line 21:
```suggestion
            data['status'] = HotWaterStatusSerializer.serialize(hot_water_dto.status)
```

This is a bit nicer than having to serialize explicitly in each api call.

src/gateway/hot_water_controller.py
```diff
@@ -0,0 +1,202 @@
...
+                if hot_water_orm is None:
+                    plugin = db.query(Plugin).where(Plugin.id == source.source_id).one()
+                    hot_water_orm = HotWater(
+                        source=Source.PLUGIN,
+                        plugin=plugin,
+                        external_id=action_payload.external_id,
+                        name=action_payload.name,
+                    )
+                    db.add(hot_water_orm)
+                    db.commit()
```
LnL7, line 85: When a new device is registered a CONFIG_CHANGE should also be published to notify the cloud.

LnL7 (approved):

